PR #9, opened 2020-10-19
Get up to date with recent developments

darosior:
This gets `revaultd` up to speed with recent changes in both practical-revault and revault_tx.

PR #10, opened 2020-10-21
Bitcoind interface

darosior:
Based on #9 , it sets up a simple interface to `bitcoind`. It's really a building block, much is still left to be done for the connection to be more reliable and to handle (most of) edgecases.

You should be able to test it on your locale node by tweaking the conf file in `test_data/` :-)

### DIFF
--- a/src/bitcoind/actions.rs
+++ b/src/bitcoind/actions.rs
@@ -80,13 +80,19 @@ pub fn wait_for_bitcoind_synced(
             // Ok, so we have some time. Let's try to avoid slowing it down by
             // spamming it with getblockchaininfo calls.
 
-            // FIXME: log here
+            if first {
+                log::info!(
+                    "Bitcoind is currently performing IBD, this is going to \
+                        take some time."
+                );
+            }
 
             // First: wait for it to gather all headers, if the current delta is
             // big enough. let's assume it won't take longer than 5min from now
             // for mainnet.
             // Then: get the number of blocks left to DL
             if first || delta > 1_000 {
+                log::info!("Waiting for bitcoind to gather enough headers..");
                 if bitcoind_config.network.eq("regtest") {
                     thread::sleep(time::Duration::from_secs(3));
                 } else {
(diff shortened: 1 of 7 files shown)


src/bitcoind/interface.rs
```diff
@@ -0,0 +1,49 @@
...
+        let client = Client::new(format!("{}", config.addr), user, pass);
+
+        Ok(BitcoinD { client })
+    }
+
+    fn make_request<'a, 'b>(
+        &self,
+        method: &'a str,
+        params: &'b [serde_json::Value],
+    ) -> Result<serde_json::Value, BitcoindError> {
```
edouardparis, line 31: I think you can change signature for:
```rust
use serde::de::DeserializeOwned;
use serde::{Deserialize, Serialize};
use serde_json;

    fn make _equest<P, T>(&self, method: &'a str, payload: P) -> Result<T, Error>
    where
        P: Serialize,
        T: DeserializeOwned,
    {
```

It will serialize the struct param and deserialize in the resource struct all using type inference

example:
https://github.com/edouardparis/opennode-rs/blob/master/opennode-client/src/client.rs#L88 
https://github.com/edouardparis/opennode-rs/blob/master/opennode-client/src/lib.rs#L51

darosior: I can, not sure i need though ? It's like `do_rpc` in the crate i use

src/bitcoind/actions.rs
```diff
@@ -0,0 +1,138 @@
...
+                    "Bitcoind is currently performing IBD, this is going to \
+                        take some time."
+                );
+            }
+
+            // First: wait for it to gather all headers, if the current delta is
+            // big enough. let's assume it won't take longer than 5min from now
+            // for mainnet.
+            // Then: get the number of blocks left to DL
+            if first || delta > 1_000 {
```
edouardparis, line 94: move 1_000, 3 and 5*60 as constants

edouardparis (approved):
fix clippy warnings

darosior:
Clippy warnings are fixed in https://github.com/re-vault/revaultd/pull/13 (even if, as discussed on MM, there are more to come :p)

src/bitcoind/actions.rs
darosior: Meh

That's kind of a hack that is going to be tinkered again, i don't even know how to name those constants.. `const DIFF_HEADERS_BLOCKS_IS_SMALL_ENOUGH = 1_000 // <- not sure about this`

darosior:
Going to merge this so we move forwards and as clippy checks are already fixed later

darosior:
ACK c599044

PR #13, opened 2020-10-28
Fixup config

darosior:
A small one on top of #12, to fixup the config (as managers are not anymore always part of the stakeholders).

Unfortunately i had to reintroduce a dev dep, see commit messages..

PR #19, opened 2020-11-05
Bitcoind main loop

darosior:
This implements the main actions on the `bitcoind` side, and as such begins the actual wallet logic. Pretty basic for now: we poll `listunspent` to stay in sync in `bitcoind`, cry blood because we have to maintain a map from address to derivation (see commit messages), and store / load deposits in the database.

We poll in the main thread, but we'll surely use a more elaborated event loop soon :)

Now going to unit test the DB and start the JSONRPC interface: hand-testing isn't going to last and we need a functional testing framework (wanted to reuse https://github.com/re-vault/revault-demo/tree/master/tests btw).


darosior:
Use FromHex https://github.com/rust-bitcoin/bitcoin_hashes/blob/master/src/hex.rs instead of my boutique function ! (Thanks Ed)

darosior:
A lot of stuff that i'd like to rework but i have to stop shaving yawks at some point. Will focus on the JSONRPC integration tomorrow.

darosior:
Rebased on top of #13 and added a test for deposits management in db.

edouardparis:
ACK 1a896f810e46f1d113fb8d9713353eb6e16b185d

PR #25, opened 2020-11-18
feat: add lib.rs for external use

edouardparis:
In order to use revaultd as an external library.


edouardparis:
Useless for now

PR #26, opened 2020-11-18
refac: move daemon_main to daemon::start

edouardparis:
In order to use revaultd as a library to start a revault daemon in Robalo, I need to remove process::exit from packages in order to have a clean way to control process and let the main function handle it.

I introduce in this change a new module: `daemon` which should manage the overall processes working in the daemon_loop.
The name collude with `revaultd`, thats why I suggest we rename `revaultd` to `revault` and  `Revaultd` to `Revault` because, this package and this struct are dedicated to the representation of the Revault protocole setup only.

### DIFF
--- a/src/bitcoind/actions.rs
+++ b/src/bitcoind/actions.rs
@@ -13,7 +13,7 @@ use revault_tx::{
 };
 
 use std::{
-    process, thread,
+    thread,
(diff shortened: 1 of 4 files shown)


edouardparis:
deprecated

PR #31, opened 2020-11-22
JSONRPC 2 server

darosior:
This implements a [JSONRPC 2](https://www.jsonrpc.org/specification) server for revaultd.

We use mio on all platforms but Windows for the transport and windows_uds on Windows. We
use jsonrpc-core for the JSONRPC messages handlers.

This also finally implements the logic of the main thread launching two separated threads,
one for the JSONRPC server and the other for the bitcoind event loop. Rust's awesome
channels abstraction is used for communicating with the main thread.

A few drive-by fixes were made (such as file permissions), and the commit history is not
*that* clean but rebasing at this point would be a huge pain and i think it's followable
enough.

As expected, Windows is **really** making our life hard. I gave up and will install a VM to
debug instead of using the CI.

Follow up work will be:
- Fixing Windows :tm:
- Implementing a basic cli (you can use netcat -U to test meanwhile)
- Implementing the agreed-upon API

### DIFF
--- a/src/bitcoind/actions.rs
+++ b/src/bitcoind/actions.rs
@@ -6,6 +6,7 @@ use crate::{
         interface::db_wallet,
     },
     revaultd::RevaultD,
+    threadmessages::ThreadMessage,
 };
 use revault_tx::{
     bitcoin::{consensus::encode, hashes::hex::FromHex, Transaction},
@@ -13,6 +14,7 @@ use revault_tx::{
 };
 
 use std::{
+    sync::{mpsc::Sender, Arc, RwLock},
     thread,
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
@@ -254,11 +256,17 @@ fn tx_from_hex(hex: &str) -> Result<Transaction, BitcoindError> {
         .map_err(|e| BitcoindError(format!("Deserializing tx hex: '{}'", e.to_string())))
 }
 
-fn poll_bitcoind(revaultd: &mut RevaultD, bitcoind: &BitcoinD) -> Result<(), BitcoindError> {
-    let (new_deposits, spent_deposits) = bitcoind.sync_deposits(&revaultd.vaults)?;
+fn poll_bitcoind(
+    revaultd: &mut Arc<RwLock<RevaultD>>,
+    bitcoind: &BitcoinD,
+) -> Result<(), BitcoindError> {
+    let (new_deposits, spent_deposits) =
+        bitcoind.sync_deposits(&revaultd.read().unwrap().vaults)?;
 
     for (outpoint, utxo) in new_deposits.into_iter() {
         let derivation_index = *revaultd
+            .read()
+            .unwrap()
             .derivation_index_map
             .get(&utxo.txo.script_pubkey)
             .ok_or_else(|| BitcoindError(format!("Unknown derivation index for: {:#?}", &utxo)))?;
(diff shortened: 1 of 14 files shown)


src/bitcoind/actions.rs
```diff
@@ -286,30 +294,34 @@ fn poll_bitcoind(revaultd: &mut RevaultD, bitcoind: &BitcoinD) -> Result<(), Bit
...
         // Mind the gap! https://www.youtube.com/watch?v=UOPyGKDQuRk
         // FIXME: of course, that's rudimentary
-        if derivation_index > revaultd.current_unused_index {
-            revaultd.current_unused_index += 1;
-            let next_addr =
-                bitcoind.addr_descriptor(&revaultd.last_deposit_address().to_string())?;
+        if derivation_index > revaultd.read().unwrap().current_unused_index {
+            revaultd.write().unwrap().current_unused_index += 1;
+            let next_addr = bitcoind
+                .addr_descriptor(&revaultd.write().unwrap().last_deposit_address().to_string())?;
```
edouardparis, line 304:
```suggestion
            let mut revaultd = revaultd.write().unwrap();
            revaultd.current_unused_index += 1;
            let next_addr = bitcoind
                .addr_descriptor(&revaultd.last_deposit_address().to_string())?;
```

edouardparis: I think it is better to do revaultd.write().unwrap() only once in the fn

src/bitcoind/actions.rs
```diff
@@ -319,9 +331,13 @@ fn poll_bitcoind(revaultd: &mut RevaultD, bitcoind: &BitcoinD) -> Result<(), Bit
             );
             // Note that it *might* have actually been confirmed during the last 30s, but it's not
             // a big deal to have it marked as unconfirmed for the next 30s..
-            db_unvault_deposit(&revaultd.db_file(), outpoint.txid.to_vec(), outpoint.vout)
-                // TODO: something like From<DbError> for BitcoindError ?
-                .map_err(|e| BitcoindError(format!("Database error: {}", e.to_string())))?;
+            db_unvault_deposit(
+                &revaultd.read().unwrap().db_file(),
```
edouardparis, line 335: db_file typically has no chance to produce a race condition, revaultd.read().unwrap() is called here only to get access to the struct. Maybe It would be cleaner to add RWLock inside of the RevaultD struct for fields like `vaults` and `derivation_index_map` and handle the lock/unlock with methods to the struct

edouardparis (commented):
cool

src/bitcoind/actions.rs
darosior: > I think it is better to do revaultd.write().unwrap() only once in the fn

While i agree it's syntactically more elegant, it would technically block the Mutex for the entire duration of the function, potentially starving other threads. No big deal *right now* but i think it's cleaner to block the Mutex as little time as possible.

src/bitcoind/actions.rs
darosior: Hmm good point. However not sure it would be `Send` then.. Also we would have to probably make it a member instead of a method.
Can we leave it for later studies of optimization (not even sure it's doable) ?

edouardparis: > Can we leave it for later studies of optimization (not even sure it's doable) ?

Yes

src/bitcoind/actions.rs
edouardparis: fair enough

edouardparis (approved):
ACK 6a7c5eb8c433ed17ffaf35657966a625ba4b23cc

PR #33, opened 2020-11-24
revault-cli: add a simple CLI

darosior:
Mainly aimed to test the JSONRPC server for now :)


Cargo.toml
```diff
@@ -4,6 +4,10 @@ version = "0.0.1"
 authors = ["Antoine Poinsot <darosior@protonmail.com>"]
 edition = "2018"
 
+[lib]
+name = "common"
+path = "src/common/lib.rs"
```
edouardparis, line 9: why not a different workspace member, importing only public modules from revaultd ?
And I am not a fan of this kind of path:
src/bin/revaultd/bitcoind/actions.rs

darosior: Pushed a new commit for a new layout which gets rid of `src/bin` in favour of `src/daemon` and `src/cli`

edouardparis (approved):
ACK 17268a9679177751b237b6bf328b588c6f66fd90

PR #36, opened 2020-11-26
Per wallet api

darosior:
Fixes https://github.com/re-vault/revaultd/issues/32 : this removes my dirty hacks :)


edouardparis (approved):
ACK dacd48c8dceab7f7dd4b4a58a9259656fbb01cfe

edouardparis:
Maybe later we will find a way to not have an Option on the wallet id

PR #48, opened 2020-12-09
contrib: dl bitcoind binary now that RCs are released

darosior:
This highly reduces the execution time of the functional tests as we don't compile bitcoind anymore


darosior:
ACK 968cbee -- trivial

PR #52, opened 2020-12-14
doc: add tx information to listvaults

edouardparis:


doc/API.md
```diff
@@ -67,16 +67,35 @@ Get an address to build a deposit transaction.
...
 
 Note that the `scriptPubKey` is implicitly known as we have the vault output Miniscript descriptor.
 **TODO** Maybe we should store and give the xpub derivation index as well ?
 
+### Vault Transaction resource
+
+| Field         | Type   | Description                                      |
+| ------------- | ------ | ------------------------------------------------ |
+| `blockheight` | int    | Block height of the transaction                  |
+| `hex`         | string | Hexadecimal format of the transaction            |
```
edouardparis, line 96: I choose the hex format of the transaction in order to the client to be able to parse it in case of needs for more tx information

doc/API.md
```diff
@@ -67,16 +67,35 @@ Get an address to build a deposit transaction.
...
-| ------------- | ------ | ------------------------------------------------------------------------------------ |
-| `amount`      | int    | Amount of the vault in satoshis                                                      |
-| `status`      | string | Status of the vault (see [vault statuses](#vault-statuses))                          |
-| `txid`        | string | Deposit txid of the vault deposit transaction                                        |
-| `vout`        | vout   | Index of the deposit output in the deposit transaction.                              |
+| Field         | Type   | Description                                                 |
+| ------------- | ------ | ----------------------------------------------------------- |
+| `amount`      | int    | Amount of the vault in satoshis                             |
+| `blockheight` | int    | Blockheight of the deposit transaction block                |
+| `created_at`  | int    | Timestamp of the deposit transaction block                  |
```
darosior, line 74: Why do you want the block's timestamp ?

doc/API.md
```diff
@@ -67,16 +67,35 @@ Get an address to build a deposit transaction.
...
+| `txid`        | string | Deposit txid of the vault deposit transaction               |
+| `updated_at`  | int    | Timestamp of the last status change                         |
+| `vout`        | int    | Index of the deposit output in the deposit transaction.     |
+
+Some fields can be added:
+
+| Field         | Type                                                      | Description                                                                |
+| ------------- | --------------------------------------------------------- | -------------------------------------------------------------------------- |
+| `final_tx`    | [Vault transaction resource](#vault-transaction-resource) | Final transaction -- not present if the vault is not spent or cancelled    |
+| `unvault_tx`  | [Vault transaction resource](#vault-transaction-resource) | Unvault transaction -- not present if vault is not unvaulting or unvaulted |
```
darosior, line 85: I wonder if it should be split instead: optional fields are IMHO non-natural. Looks like you are describing executed transactions here so maybe a `listconchaintransactions [vault_id]` call ?

Wouldn't return anything for the unconfirmed vaults, the deposit for the funded, the unvault for the unvaulted, the spend too if spent, or the cancel.

doc/API.md
```diff
@@ -67,16 +67,35 @@ Get an address to build a deposit transaction.
...
+
 
 Note that the `scriptPubKey` is implicitly known as we have the vault output Miniscript descriptor.
 **TODO** Maybe we should store and give the xpub derivation index as well ?
 
+### Vault Transaction resource
```
darosior, line 91: So, yeah, i think this pattern confirms my comment above :)

doc/API.md
darosior: Right, so the `txid` is redundant. Also i think the timestamp should be the reception-time rather than block time

darosior (commented):
Some nice suggestions! Argued a bit in favor of my shed for this bike :)

doc/API.md
edouardparis: ACK reception time

doc/API.md
```diff
@@ -96,6 +107,26 @@ either `status` or deposit `txids`.
...
+
+| Parameter        | Type    | Description                                     |
+| ---------------- | ------- | ----------------------------------------------- |
+| `txid`           | string  | Deposit txid of the vault                       |
+
+#### Response
+
+| Field         | Type                                                      | Description                                                                |
+| ------------- | --------------------------------------------------------- | -------------------------------------------------------------------------- |
+| `cancel_tx`   | [Vault transaction resource](#vault-transaction-resource) | Cancel transaction -- not present if the vault is cancelling or cancelled  |
```
darosior, line 125: nit: You missed a "not" here

darosior (approved):
ACK d6238e8438f623c0f6b76a16ac60f4a904966eed

darosior:
Hmm there is def something going on with the CI.

darosior:
Anyways,

ACK 9a13340

doc/API.md
```diff
@@ -9,6 +9,7 @@ Note that all addresses are bech32-encoded *version 0* native Segwit `scriptPubK
 | ----------------------------------------- | ---------------------------------------------------- |
 | [`getinfo`](#getinfo)                     | Display general information                          |
 | [`listvaults`](#listvaults)               | Display a paginated list of vaults                   |
+| [`getonchaintxs`](#getonchaintxs)         | Retrieve the vault on chain transactions             |
```
darosior, line 12: Actually, since we don't have a call to fetch all the transactions it could be a `getvaulttxs [vault_outpoint]` and the fact that they are or not onchain is implied by their status in `listvaults`.. Meh

doc/API.md
```diff
@@ -96,6 +107,26 @@ either `status` or deposit `txids`.
...
+### `getonchaintxs`
+
+The `getonchaintxs` RPC Command retrieves the on chain transactions of a
+vault with the given deposit `txid`.
+
+#### Request
+
+| Parameter        | Type    | Description                                     |
+| ---------------- | ------- | ----------------------------------------------- |
+| `txid`           | string  | Deposit txid of the vault                       |
```
darosior, line 119: As told on MM we need to update all those to take outpoints instead.

PR #53, opened 2020-12-15
Implement `getrevocationtxs`

darosior:
This implements the `getrevocationtxs` RPC, as well as a few database interface stuff and fixes the documentation (~~but not yet the implem!~~ now implemented) for the vault identifier.

~~I'll add some tests tomorrow but you can already give it a look, i tested it manually :tm:~~ Added a functional test

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -130,7 +130,8 @@ vault with the given deposit `txid`.
 ### `getrevocationtxs`
 
 The `getrevocationtxs` RPC Command builds and returns the revocation transactions
-corresponding to a given vault.
+corresponding to a given vault. The call will fail if the `outpoint` does not refer to a
+known and confirmed ([`funded`](#vault-statuses)) vault.
 
 #### Request
 
@@ -235,7 +236,7 @@ amount of the output.
   | +------sig---------> |                          |
   |                      |                          |
   | <+sign unvault_emer+ |                          |
-  | +------------------> |                          |
+  | +------sig---------> |                          |
   |                      | +---revocationtxs------> |
   |                      |                          |
   +                      | +--+listvaults secure+-> |  // check if the watchtowers has the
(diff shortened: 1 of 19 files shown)


darosior:
So the next suspect will be https://github.com/re-vault/revaultd/blob/master/doc/API.md#getonchaintxs, which will need the db interface anyways so dead code won't last :)

src/daemon/database/interface.rs
```diff
@@ -184,65 +192,197 @@ pub struct DbVault {
     pub derivation_index: ChildNumber,
 }
 
+fn vault_from_row(row: &Row) -> rusqlite::Result<DbVault> {
```
edouardparis, line 195: May I suggest:

```rust
impl TryFrom<&Row<'_>> for DbVault {
    type Error = rusqlite::Error;
    fn try_from(row: &Row) -> Result<Self, Self::Error> {
```
then 
```diff
-         vault_from_row,
+        |row: &Row| row.try_into(),
```

doc/API.md
```diff
@@ -92,14 +92,14 @@ Note that the `scriptPubKey` is implicitly known as we have the vault output Min
...
 
 #### Request
 
-| Parameter | Type         | Description                                                                                     |
-| --------- | ------------ | ----------------------------------------------------------------------------------------------- |
-| `status`  | string array | Vault status -- optional, see [vault statuses](#vault-statuses) for possible values             |
-| `txids`   | string array | Vault IDs -- optional, filter the list with the given vault IDs                                 |
+| Parameter   | Type         | Description                                                                                     |
+| ----------- | ------------ | ----------------------------------------------------------------------------------------------- |
+| `status`    | string array | Vault status -- optional, see [vault statuses](#vault-statuses) for possible values             |
```
edouardparis, line 101: nit: order fields

src/daemon/database/interface.rs
darosior: You may :p 
Addressed!

doc/API.md
darosior: Addressed..

darosior:
Ok, i think it's now gtg

edouardparis (approved):
ACK 4c7c1e637ec7c579491bb69ea4a4ab5d3c138eaa

PR #55, opened 2020-12-18
bitcoind: update to latest (and stable) rust-jsonrpc

darosior:
This PR is not yet ready as it's relying on Steven Roose's branch but they are close to release, see https://github.com/apoelstra/rust-jsonrpc/pull/39 .

PR #56, opened 2020-12-20
Listtransactions

darosior:
This implements the `listtransactions` RPC call, which takes a vault deposit as input and outputs:
- The deposit transaction, always as `hex` along with its reception time (`received_at`). Optionally with a `blockheight` entry if it's confirmed.
- The pre-signed transactions (unvault, cancel, both emergency) as a `psbt` until it's fully signed. Then at `hex` along with its reception time (`received_at`), ie the time our `bitcoind` noticed it was broadcast. If it's confirmed, a `blockheight` entry is present as well.
- A spend transaction, only if a spend is ongoing. (same fields as above)

Based on #55

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -128,6 +128,7 @@ either `status` or deposit `outpoints`.
 | `blockheight` | int    | Height of the block containing the transaction, `0` if unconfirmed         |
 | `psbt`        | string | base64-serialized BIP174 format of the transaction, if not fully-signed    |
 | `hex`         | string | Hexadecimal of the network-serialized transaction, if fully-signed         |
+| `received_at` | int    | Transaction reception time as UNIX epoch timestamp                         |
 
 
 ### `getrevocationtxs`
--- a/doc/API.md
+++ b/doc/API.md
@@ -8,11 +8,11 @@ Note that all addresses are bech32-encoded *version 0* native Segwit `scriptPubK
 | Command                                   | Description                                          |
 | ----------------------------------------- | ---------------------------------------------------- |
 | [`getinfo`](#getinfo)                     | Display general information                          |
-| [`listvaults`](#listvaults)               | Display a paginated list of vaults                   |
-| [`getonchaintxs`](#getonchaintxs)         | Retrieve the vault on chain transactions             |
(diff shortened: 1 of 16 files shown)


doc/API.md
```diff
@@ -107,25 +100,36 @@ either `status` or deposit `outpoints`.
...
-### `getonchaintxs`
 
-The `getonchaintxs` RPC Command retrieves the on chain transactions of a
-vault with the given deposit `outpoint`.
+### `listtransactions`
 
-#### Request
+| Parameter   | Type         | Description                                                                                     |
+| ----------- | ------------ | ----------------------------------------------------------------------------------------------- |
+| `outpoints` | string array | Vault IDs -- optional, filter the list with the given vault Outpoints                           |
```
edouardparis, line 108: in this PR, the listtransactions command as an outpoint in its params which is required

darosior: Right, the doc is wrong thanks :+1:

doc/API.md
```diff
@@ -107,25 +100,36 @@ either `status` or deposit `outpoints`.
...
-| Parameter        | Type    | Description                                     |
-| ---------------- | ------- | ----------------------------------------------- |
-| `outpoint`       | string  | Deposit outpoint of the vault                   |
+// FIXME: we could eventually also take an array of transaction types here.
 
-#### Response
+### Response
+
+| Field               | Type                                                 | Description                                                            |
+| ------------------- | ---------------------------------------------------- | ---------------------------------------------------------------------  |
```
edouardparis, line 115: add vault_outpoint

doc/API.md
edouardparis: I think the doc is right :)

darosior: Lol, it was at the time of writing but now that we agreed it effectively is!

darosior:
Added a commit which makes `listtransactions` take an array of outpoints, and add an `outpoint` field to each "transactions resource"

edouardparis (approved):
light ACK 47c972ea11aee3efc0d1f681b6a427a1064cc1f6
I need to test it but I agree to merge it

PR #60, opened 2020-12-21
Revocationtxs

### LINKED ISSUE #57: db: store psbt as a blob
It's currently stored as a base64 string... Which makes no sense

darosior:
This (partially) implements the `revocationtxs` call. It's not functional yet, but the only thing left is the signature fetching from the sync server.

On top of #56 
Fixes  #57


darosior:
Rebased after #56 changes (first commit is 0c0c8a5)

darosior:
Rebased on #62 which appears to solve flakiness

darosior:
Merged in #63

PR #61, opened 2020-12-21
rpc: server: don't read garbage on WouldBlock

darosior:


darosior:
ACK f7db4ef -- trivial bugfix

edouardparis (approved):
Tried several time, the problem did not appear.
At least if this patch does not solve it, it does not break something.
ACK df499863680e836d98ab7e5e15e109df06ef924e

PR #63, opened 2020-12-24
Cleanups and refactoring

darosior:
This is an effort to re-organize our modules (still undecided wrt the weird actions and interfaces in each of bitcoind and database), add some documentation, and make our code overall more readable.

No new functionality here (but some fixes and tests), but i think it would now look nicer to an external reader (still not perfect but hey, that's a first step and the pull is becoming huge!).

Based on #60


edouardparis (approved):
ACK a9fbe444fa92462f32d2f1dc2d34b9f8de83447e
Maybe split now the `control::handle_rpc_messages` with different handlers and a router

darosior:
Agree, can definitely be done in a follow-up.

PR #66, opened 2021-01-07
Update jsonrpc to stable

darosior:
Finally!

PR #69, opened 2021-01-21
Small refactoring before sig sharing

darosior:
I split that from my sig sharing branch to avoid a monster PR. ~~This is based on #66 , which i need to figure out how it broke Windows.. :sob:~~ EDIT: fixed and merged #66

This implements part of https://github.com/re-vault/revaultd/issues/54, i wonder how best is to tackle the last field you ask. Maybe we could add a timestamp to the vault table, but i wonder if it'd be consistent with the `created_at` from `bitcoind`.

### DIFF
--- a/contrib/ci-functional-tests.sh
+++ b/contrib/ci-functional-tests.sh
@@ -4,10 +4,10 @@ set -xe
 cargo build --release
 
 # Download the bitcoind binary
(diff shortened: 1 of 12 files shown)


edouardparis (approved):
ACK 5a69cd6905ea852a8ff8a82d5ac0a44e5c27cc6f

PR #72, opened 2021-01-31
rpc: don't overwrite buffer when reading from stream

darosior:
Unfortunately i don't have the time to add a test rn, but we should have hardened that..

### DIFF
--- a/src/daemon/jsonrpc/server.rs
+++ b/src/daemon/jsonrpc/server.rs
@@ -8,7 +8,7 @@ use crate::{
 
 use std::{
     collections::{HashMap, VecDeque},
-    io::{self, Read, Write},
+    io::{self, Write},
     path::PathBuf,
     sync::{mpsc::Sender, Arc, RwLock},
 };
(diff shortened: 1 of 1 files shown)


edouardparis:
Please change fn arguments for:
```diff
-fn read_bytes_from_stream(mut stream: &UnixStream) -> Result<Option<Vec<u8>>, io::Error> {
+fn read_bytes_from_stream(stream: &mut dyn io::Read) -> Result<Option<Vec<u8>>, io::Error> {
```
and add unit test:

```rust
    use super::read_bytes_from_stream;
    use std::io::Cursor;
    #[test]
    fn read_bytes_from_stream_test() {
        let data = vec![1; 522];
        let mut stream = Cursor::new(data.clone());
        let res = read_bytes_from_stream(&mut stream);
        assert_eq!(res.unwrap().unwrap(), data);
    }
```

darosior:
Thanks! Will improve over the unit test now that you mainly wrote it :)

darosior:
Finally got to adding it. We'll improve it later, for now i need to PR the signature sharing :)

darosior:
ACK 3403758

PR #76, opened 2021-02-08
rpc: add `update_at` field for `listvaults`

darosior:
This a rebase of Edouard Paris' patch which fixes #54 on top of #75 .

PR #78, opened 2021-02-10
Incremental parsing for the JSONRPC server

darosior:
This builds on top of #76 .

This implements a read cache per connection so that commands can be split across reads. While at it i also implemented a pretty naive parallelization of command handlers, as it was a pain to use the CLI at the same time as the GUI.

All of this puts some more load on us but will make the interface way more robust --at least i hope so :).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -740,7 +740,7 @@ dependencies = [
 [[package]]
 name = "revault_tx"
 version = "0.0.1"
-source = "git+https://github.com/re-vault/revault_tx#7a13866cc09e0ee7e582f5f50f7ab5d6702f3723"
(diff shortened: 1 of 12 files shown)


darosior:
Rebased

darosior:
Rebased and dropped the max number of threads to 4.

darosior:
CI was a spurious error, tried to unflake it with the latest commit.

darosior:
Added a last commit after https://github.com/re-vault/revault_tx/pull/52 merge. It really only is a word-changed (almost entirely `sed`-ed), `--word-diff` can help here for the review.

edouardparis (approved):
ACK 4b58912ec9171d7d386c5cfaa252739e6ad71562

PR #84, opened 2021-02-15
Remove Emergency address from managers

darosior:
This updates to latest revault_tx and revault_net to finally be able to prevent the manager to create Emergency transactions.
Note that in `test_sig_sharing` was *not* modified to only check stakehodlers pass to `SECURED` state: managers will implicitly pass to the SECURED state as all their presigned transactions are now signed (their Emergency transactions were never created in the first place).
The SECURED state does not make much sense for stakeholders but it will ease the passage to ACTIVE later on.

## Warning

This broke the `listtransactions` API as a result of not allowing manager to be able to create the transactions at all. `listtransactions` is fucked up anyways and should be burnt with fire, but still, sorry for the breakage Ed!


edouardparis (commented):
Github does not make it clear what are the change you did in `coordinatord`

darosior:
> Github does not make it clear what are the change you did in coordinatord

Yes, sorry about that: i god-mode merged the changes in https://github.com/re-vault/coordinatord/pull/5 , so this is just an update to coordinatord's master.

src/common/config.rs
```diff
@@ -1,8 +1,56 @@
...
 
-use revault_tx::bitcoin::{util::bip32, Address, Network, PublicKey};
+use revault_net::noise::PublicKey as NoisePubKey;
+use revault_tx::bitcoin::{hashes::hex::FromHex, util::bip32, Address, Network, PublicKey};
 
-use serde::Deserialize;
+use serde::{de, Deserialize};
+
+#[derive(Debug, Clone)]
+pub struct NoisePubkeyHex {
```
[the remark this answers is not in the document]
darosior: Yea, but it's just really a hack. I'm considering proposing a patch to sodiumoxide to implement a string visitor in their serde implem
Let's keep it as is for now

src/common/config.rs
```diff
@@ -26,6 +74,7 @@ pub struct WatchtowerConfig {
 pub struct StakeholderConfig {
     pub xpub: bip32::ExtendedPubKey,
     pub watchtowers: Vec<WatchtowerConfig>,
+    pub emergency_address: Address,
```
edouardparis, line 77: nit: Change address for revault_tx::EmergencyAddress using a serde Deserializer
example: https://github.com/re-vault/revault-gui/blob/master/src/revaultd/model.rs#L131

src/daemon/revaultd.rs
```diff
@@ -344,7 +342,11 @@ impl RevaultD {
             config.unvault_csv,
         )?;
         let cpfp_descriptor = cpfp_descriptor(managers_pubkeys)?;
-        let emergency_address = EmergencyAddress::from(config.emergency_address)?;
+        let emergency_address = if let Some(conf) = config.stakeholder_config {
+            Some(EmergencyAddress::from(conf.emergency_address)?)
```
edouardparis, line 346: I think it is better to throw an error during the deserialization of the config than in the daemon. see https://github.com/re-vault/revaultd/pull/84/files#r576995072

darosior: I agree, i was being lazy :) Will amend.

src/common/config.rs
darosior: Added a commit upstream for this, as we may use it in other projects too.  See https://github.com/re-vault/revault_tx/pull/58#issuecomment-781419432

darosior:
Updated upstream, we now parse the Emergency address from config! :tada:

darosior:
Update to revault_tx master

edouardparis (approved):
ACK 7a69884eef00cdd1a9df70b3e0ca3574052c2e8d

Issue #82, opened 2021-02-12
Fix bitcoind lag


darosior:
https://github.com/re-vault/revaultd/commit/a4690dfc334a3daaf71885b4e2b60afc358479a6 was purely dumb, but i don't want to get back to 100% CPU..

edouardparis:
bitcoind_main_loop could use https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.iter instead of try_recv in order to block the thread, and A ticker could send to bitcoin_main_loop a tick in order to check the blockchain tip.

darosior:
Yes, i was thinking either this or just spawn the poller in a separate thread.

edouardparis:
A mix of both would be cool :+1: 
thread with receiver should use `iter`

darosior:
Yeah, the poller was the reason we could not use an iterator in the first place

darosior:
Ok, lest fix it post https://github.com/re-vault/revaultd/issues/73 .

PR #86, opened 2021-02-17
Unvault signatures sharing and fetching

darosior:
This builds on top of #84 to implement `getunvaulttx` and `unvaulttx` in order for stakeholders to exchange Unvault Transaction signatures, activating vaults.

This largely re-uses the logic from revocation sigs sharing, see commit messages for details.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -141,7 +141,7 @@ either `status` or deposit `outpoints`.
 
 ### `getrevocationtxs`
 
-The `getrevocationtxs` RPC Command builds and returns the revocation transactions
+The `getrevocationtxs` RPC Command builds and returns the (unsigned) revocation transactions
 corresponding to a given vault. The call will fail if the `outpoint` does not refer to a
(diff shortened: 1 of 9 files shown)


darosior:
Addressed @edouardparis feedback from #82

darosior:
Rebased after #84 merge

edouardparis (commented):
what if we store the signature and failed to share it to coordinator and the daemon stopped for some reason ? We may need a worker which share again unshared txs no or the client will do the process of signing again and overwrite the tx in db ?

edouardparis:
ACK 11c7026f5434e5a9d497b6f6dec347e458b50303

edouardparis (approved):
ACK 11c7026f5434e5a9d497b6f6dec347e458b50303

darosior:
You are right (and it was already the case for `revocationtxs`), opened https://github.com/re-vault/revaultd/issues/96 so we don't lose track of this edge case :)

PR #97, opened 2021-02-23
fix(jsonrpc/api): missing vault_outpoint

edouardparis:
A vault can share the same deposit transaction with
an other vault. The client need to have information
about which transactions are linked to the requested
outpoint.


darosior:
Good catch

darosior:
ACK ebd7981

PR #99, opened 2021-02-24
Basic handling of reorgs

darosior:
This turned out to have even more edge cases than i expected. Went for a bulk (rescan everything on reorg) method instead of a more smarter-and-footguny (keep a header chain and rescan based on deposit blockheight) one. See commit messages for more comprehensive details.

Note that this is only a partial implementation, as we'll eventually have to deal with all the remaining transactions.

Given the slippery slop that it is by itself i separated the functional test in a previous commit for reviewers to be able to evaluate the edge cases coverage. A slippery slop it's even more with the concurrency (RPC thread, signature fetcher) but for those ones the race is so intense it can't be tested reproducibly.

### DIFF
--- a/tests/serializations.py
+++ b/tests/serializations.py
@@ -25,14 +25,18 @@ import hashlib
 import copy
 import base64
 
+
 def sha256(s):
-    return hashlib.new('sha256', s).digest()
+    return hashlib.new("sha256", s).digest()
+
 
 def ripemd160(s):
-    return hashlib.new('ripemd160', s).digest()
+    return hashlib.new("ripemd160", s).digest()
+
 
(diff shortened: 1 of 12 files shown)


tests/serializations.py
```diff
@@ -25,15 +25,19 @@
 import copy
```
edouardparis, line 25: looks like Blake did a big job

darosior: Glad you like it, it's increasing the diff and the blame depth but worth it imho

edouardparis (approved):
ACK d16c90a1a35bdd1bdb294c30af1f342697c2ed55

PR #111, opened 2021-03-11
Add missing `blockheight` and `received_at` fields to `listvaults`

### LINKED ISSUE #98: Add missing `blockheight` and `received_at` fields to `listvaults`
That's what the doc specifies but we don't actually do it !

danielabrozzoni:
Closes #98

### DIFF
--- a/src/daemon/database/actions.rs
+++ b/src/daemon/database/actions.rs
@@ -258,13 +258,14 @@ pub fn db_insert_new_unconfirmed_vault(
     deposit_outpoint: &OutPoint,
     amount: &Amount,
     derivation_index: ChildNumber,
+    received_at: u32,
 ) -> Result<(), DatabaseError> {
     db_exec(db_path, |tx| {
         let derivation_index: u32 = derivation_index.into();
         tx.execute(
             "INSERT INTO vaults (wallet_id, status, blockheight, deposit_txid, \
-             deposit_vout, amount, derivation_index, updated_at) \
-             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, strftime('%s','now'))",
+             deposit_vout, amount, derivation_index, received_at, updated_at) \
+             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, ?9)",
             params![
                 wallet_id,
                 *status as u32,
@@ -273,6 +274,8 @@ pub fn db_insert_new_unconfirmed_vault(
                 deposit_outpoint.vout,
                 amount_to_i64(amount),
                 derivation_index,
+                received_at,
+                received_at,
             ],
         )
         .map_err(|e| DatabaseError(format!("Inserting vault: {}", e.to_string())))?;
@@ -576,6 +579,7 @@ mod test {
         )
         .unwrap();
         let amount = Amount::from_sat(123456);
+        let received_at = 1615297315;
         let derivation_index = ChildNumber::from(3);
         db_insert_new_unconfirmed_vault(
             &db_path,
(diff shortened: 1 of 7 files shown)


darosior (commented):
I conceptually found weird to add a column to the DB for an information we have on bitcoind, but hey practically it'd mean having a call to bitcoind per vault which is rather ugly.

Why only have `received_at` for confirmed vaults though ? `updated_at` is for all statuses so i'd say go with setting it in `db_insert_new_unconfirmed_vault` ?

danielabrozzoni:
Are there any edge cases where `timereceived` in `gettransaction` could change for an unconfirmed tx? I don't think so but I'm quite unsure

darosior:
No, in bitcoind the `timereceived` is set at reception and never changed afterward. `time` may be changed though.

darosior:
Arg the functional tests are deadlocking somewhere..

src/daemon/database/schema.rs
```diff
@@ -46,6 +46,7 @@ CREATE TABLE vaults (
     wallet_id INTEGER NOT NULL,
     status INTEGER NOT NULL,
     blockheight INTEGER NOT NULL,
+    received_at INTEGER NOT NULL,
```
darosior, line 49: Bikeshedding but just to cure my OCD i feel like it should be just before `updated_at`

src/daemon/database/actions.rs
```diff
@@ -255,20 +255,22 @@ pub fn db_insert_new_unconfirmed_vault(
...
 ) -> Result<(), DatabaseError> {
     db_exec(db_path, |tx| {
         let derivation_index: u32 = derivation_index.into();
         tx.execute(
-            "INSERT INTO vaults (wallet_id, status, blockheight, deposit_txid, \
-             deposit_vout, amount, derivation_index, updated_at) \
-             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, strftime('%s','now'))",
+            "INSERT INTO vaults (wallet_id, status, blockheight, received_at, \
+             deposit_txid, deposit_vout, amount, derivation_index, updated_at) \
+             VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7, ?8, strftime('%s','now'))",
```
darosior, line 268: I think this is super weird that
```
received_at == updated_at
```
does not hold for unconfirmed vaults and that it could be confusing once displayed on the UI. What about setting `updated_at` to the value of `received_at` here ?

darosior (commented):
Bikeshedding..

darosior:
ACK a82d319

Tested with a small sanity blackbox test:
```patch
index e79826b..970ab74 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -37,6 +37,7 @@ def test_listvaults(revaultd_manager, bitcoind):
     assert vault_list[0]["amount"] == amount_sent * 10 ** 8
     assert vault_list[0]["address"] == addr
     assert vault_list[0]["derivation_index"] == 0
+    assert vault_list[0]["updated_at"] == vault_list[0]["received_at"]
 
     # Generate 5 blocks, it is still unconfirmed
     bitcoind.generate_block(5)
@@ -47,7 +48,9 @@ def test_listvaults(revaultd_manager, bitcoind):
     # 1 more block will get it confirmed
     bitcoind.generate_block(1)
     revaultd_manager.wait_for_log(f"Vault at .*{txid}.* is now confirmed")
-    assert revaultd_manager.rpc.call("listvaults")["vaults"][0]["status"] == "funded"
+    vault = revaultd_manager.rpc.call("listvaults")["vaults"][0]
+    assert vault["status"] == "funded"
+    assert vault["updated_at"] > vault["received_at"]
 
     # Of course, it persists across restarts.
     revaultd_manager.rpc.call("stop")
```

PR #114, opened 2021-03-15
Add total number of vaults to `getinfo`

### LINKED ISSUE #101: Add total number of vaults to `getinfo`

danielabrozzoni:
Closes #101

### DIFF
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -495,7 +495,18 @@ pub fn handle_rpc_messages(
                     ..
                 } = db_tip(&db_path)?;
 
-                response_tx.send((network.to_string(), blockheight, progress))?;
+                let number_of_vaults = listvaults_from_db(&revaultd.read().unwrap(), None, None)?
+                    .iter()
+                    .filter(|l| {
+                        l.status != VaultStatus::Spent
+                            && l.status != VaultStatus::Canceled
+                            && l.status != VaultStatus::Unvaulted
+                            && l.status != VaultStatus::EmergencyVaulted
+                    })
+                    .collect::<Vec<_>>()
+                    .len();
+
+                response_tx.send((network.to_string(), blockheight, progress, number_of_vaults))?;
             }
             RpcMessageIn::ListVaults((statuses, outpoints), response_tx) => {
                 log::trace!("Got listvaults from RPC thread");
--- a/src/daemon/jsonrpc/api.rs
+++ b/src/daemon/jsonrpc/api.rs
@@ -202,7 +202,7 @@ impl RpcApi for RpcImpl {
             meta.tx.send(RpcMessageIn::GetInfo(response_tx)),
             "Sending 'getinfo' to main thread"
         );
-        let (net, height, progress) = assume_ok!(
+        let (net, height, progress, number_of_vaults) = assume_ok!(
             response_rx.recv(),
             "Receiving 'getinfo' result from main thread"
         );
@@ -212,6 +212,7 @@ impl RpcApi for RpcImpl {
             "network": net,
             "blockheight": height,
             "sync": progress,
+            "vaults": number_of_vaults,
         }))
     }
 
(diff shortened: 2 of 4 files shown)


edouardparis:
I am not sure we want to count spent vaults

darosior:
Why not? Though i agree a rationale for the change would help here :)

danielabrozzoni:
Right, didn't though of that, I agree that showing *every* vault that ever happened is quite useless. Since we want to show that on the UI I would count only the active vaults? What do you think of that?

darosior:
Maybe all the confirmed ones?

danielabrozzoni:
(Relevant for https://github.com/revault/revault-gui/issues/30)
How do we want to display this to the user? `Confirmed vaults`? Also, what is exactly a confirmed vault, one either `funded`, `secured`, `active`?

darosior:
A vault [is marked as `funded` once considered confirmed](https://github.com/revault/revaultd/blob/master/doc/API.md#vault-statuses). We consider a vault to be confirmed after its deposit tx is N blocks deep in the chain, N is currently hardcoded to 6 but it'll change

danielabrozzoni:
Cool, updated to `number_of_funded_vaults` (you can suggest a better name 😅 ). The variable is an usize but I'm not sure it makes sense to return an `usize` to the user, maybe a `u32` is better?

src/daemon/control.rs
```diff
@@ -493,7 +493,19 @@ pub fn handle_rpc_messages(
                     ..
                 } = db_tip(&db_path)?;
 
-                response_tx.send((network.to_string(), blockheight, progress))?;
+                let number_of_funded_vaults =
+                    listvaults_from_db(&revaultd.read().unwrap(), None, None)?
+                        .iter()
+                        .filter(|l| l.status == VaultStatus::Funded)
```
darosior, line 499: This would discard all vaults that are not **only** funded (ie active, unvaulting, w/e). I suggest that we go with `l.status != VaultStatus::Spent` and just rename the field `vaults` in the `getinfo` result.

danielabrozzoni: Ah, ok, thought your comment meant to count only the funded :) I agree, counting the non spent makes more sense

darosior:
ACK d6ab898

It's trivial but a sanity check is very welcome anyways (as it's trivial to implement too). May i suggest
```patch
index e79826b..ccbb8b7 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -14,6 +14,7 @@ def test_getinfo(revaultd_manager, bitcoind):
     assert res["network"] == "regtest"
     assert res["sync"] == 1.0
     assert res["version"] == "0.0.2"
+    assert res["vaults"] == 0
 
     wait_for(lambda: revaultd_manager.rpc.call("getinfo")["blockheight"] > 0)
     height = revaultd_manager.rpc.call("getinfo")["blockheight"]
@@ -37,6 +38,7 @@ def test_listvaults(revaultd_manager, bitcoind):
     assert vault_list[0]["amount"] == amount_sent * 10 ** 8
     assert vault_list[0]["address"] == addr
     assert vault_list[0]["derivation_index"] == 0
+    assert revaultd_manager.rpc.call("getinfo")["vaults"] == 1
 
     # Generate 5 blocks, it is still unconfirmed
     bitcoind.generate_block(5)
@@ -99,8 +101,8 @@ def test_largewallets(revaultd_stakeholder, bitcoind):
             txids.append(bitcoind.rpc.sendtoaddress(addr, amount))
         bitcoind.generate_block(6, txids)
 
-    revaultd_stakeholder.rpc.getinfo()
-    revaultd_stakeholder.rpc.listvaults()
+    wait_for(lambda: revaultd_stakeholder.rpc.getinfo()["vaults"] == 10 * 100)
+    assert len(revaultd_stakeholder.rpc.listvaults()["vaults"]) == 10 * 100
 
 
```
which i used for testing

danielabrozzoni:
Updated it, thanks :)

darosior:
re-ACK ef066cd

Issue #87, opened 2021-02-21
Organisation: a new `control` module


darosior:
`control.rs` is getting large and we could separate "domains" of control better (eg signature sharing / checking, bitcoind interface, db interface, ..)

PR #129, opened 2021-03-23
Spending process: Cosigning server + Coordinator communication, and Unvault + Spend broadcast

darosior:
This implements the spending process on top of #125 .

We add a `spend_transactions` table to the database in order to keep track of PSBTs. A bridge `spend_inputs` table is added as well to join with the Unvault transaction in the `presigned_transactions` table.

This allows to offer to the caller a way to:
1. Create a Spend transaction (previous work, already in master)
2. Store or update a Spend transaction (`updatespendtx`)
3. Delete a stored Spend transaction (`delspendtx`)
4. List the stored Spend transaction along with the vault deposit outpoints they ultimately spend (`listspendtxs`)
5. Trigger the Spending Process by
5.a Checking all managers signed
5.b Requesting the signatures from the cosigning servers
5.c Announcing the Spend transaction to the Coordinator
5.d Broadcasting the Unvault transaction(s)
5.e Later broadcasting the Spend transaction

We also integrate the Cosigning Server in the functional test framework: https://github.com/revault/cosignerd .

This does not contain proper state tracking which will be done in a followup.

Fixes #118

PR #133, opened 2021-03-27
 Run the JSONRPC API in the main thread

### LINKED ISSUE #115: Run the JSONRPC API in the main thread
This is unnecessary complexity: all the main thread is doing is listening for commands, and i don't expect it to change soon

### LINKED ISSUE #116: Return JSONRPC errors rather than crashing during command handling
I was against it but now i tend to think it's cleaner once we have actual users.. Anyways needs #115 first

danielabrozzoni:
This PR is based on #129 and closes #115, closes #116 

Closes https://github.com/revault/revaultd/issues/87


danielabrozzoni:
Still TODO: update the documentation to reflect the fact that jsonrpc thread doesn't exist anymore

src/daemon/control.rs
```diff
@@ -614,784 +604,35 @@ fn announce_spend_transaction(
...
+        sigfetcher_thread: JoinHandle<()>,
+    ) -> Self {
+        Self {
+            revaultd,
+            db_path,
+            network,
+            bitcoind_tx,
+            bitcoind_thread: Arc::new(RwLock::new(Some(bitcoind_thread))),
+            sigfetcher_tx,
+            sigfetcher_thread: Arc::new(RwLock::new(Some(sigfetcher_thread))),
```
darosior, line 635: Why having a constructor if fields are public ?

src/daemon/control.rs
```diff
@@ -614,784 +604,35 @@ fn announce_spend_transaction(
...
-                        response_tx.send(Err(RpcControlError::UnvaultBroadcast(e.to_string())))?;
-                        continue;
-                    }
-                }
-                db_mark_broadcastable_spend(&db_path, &spend_txid)?;
+#[derive(Clone)]
+pub struct RpcUtils {
+    pub revaultd: Arc<RwLock<RevaultD>>,
+    pub db_path: PathBuf,
+    pub network: Network,
```
darosior, line 611: These are a duplicate from RevaultD, i think you should be retrieving them from it instead

src/daemon/jsonrpc/api.rs
```diff
@@ -219,26 +241,71 @@ impl RpcApi for RpcImpl {
     type Metadata = JsonRpcMetaData;
 
     fn stop(&self, meta: JsonRpcMetaData) -> jsonrpc_core::Result<()> {
+        meta.rpc_utils
+            .bitcoind_tx
+            .send(BitcoindMessageOut::Shutdown)
+            .map_err(|_| JsonRpcError::internal_error())?;
```
darosior, line 247: Hmm. I acked using JSONRPC's internal error but we need the message. The `jsonrpc_core::Error` struct is public, therefore maybe we could have a helper creating it out of a string message?

src/daemon/jsonrpc/api.rs
```diff
@@ -544,19 +1032,49 @@ impl RpcApi for RpcImpl {
         spend_tx: SpendTransaction,
     ) -> jsonrpc_core::Result<serde_json::Value> {
         manager_only!(meta);
+        log::trace!("Got 'updatespendtx' request from RPC thread");
```
darosior, line 1035: Some coppypasta leftover :)

src/daemon/control.rs
```diff
@@ -9,23 +9,17 @@
...
     database::{
-        actions::{
-            db_delete_spend, db_insert_spend, db_mark_broadcastable_spend, db_update_presigned_tx,
-            db_update_spend,
-        },
         interface::{
-            db_cancel_transaction, db_emer_transaction, db_list_spends, db_spend_transaction,
-            db_tip, db_unvault_emer_transaction, db_unvault_transaction, db_vault_by_deposit,
-            db_vault_by_unvault_txid, db_vaults, db_vaults_from_spend,
+            db_cancel_transaction, db_emer_transaction, db_unvault_emer_transaction,
```
darosior, line 13: In commit 9ee70ef375f4282873c4dc4f0ffd5df759c75a3e, you should also remove the reference to the JSONRPC thread in main.rs, and i think there are elsewhere too

darosior (commented):
Nice, definitely looks like the Right Thing :tm:

A few comments i had from a first skim through it, working on fixing the JSONRPC server test now.

src/daemon/control.rs
danielabrozzoni: I honestly don't know what the convention is, but using constructors only when fields are private seems reasonable. I'll update it :)

darosior:
Hmm, you also need to tweak the shutdown mechanism (try to call stop on a regtest daemon)

src/daemon/sigfetcher.rs
```diff
@@ -251,17 +251,17 @@ pub fn signature_fetcher_loop(
...
+        // If enough time has elapsed, poll the sigs
+        if elapsed >= poll_interval {
+            // This will ignore emergency transactions if we are manager-only
+            let txs = db_transactions_sig_missing(&revaultd.read().unwrap().db_file())?;
+            log::trace!("Fetching transactions for {:#?}", txs);
+            fetch_all_signatures(&revaultd.read().unwrap(), txs).unwrap_or_else(|e| {
+                log::warn!("Error while fetching signatures: '{}'", e);
+            });
+
+            last_poll = time::Instant::now();
```
darosior, line 264: Huh, this will be a CPU hog: we need to be smarter. Maybe just sleep for half a second. Or use `recv_timeout` with the `clone` hack to avoid the panic..

danielabrozzoni:
Tests are continuously printing `Process died while waiting for logs`. On it. :(

darosior:
Ok, fixed the jsonrpc server test and squashed it

danielabrozzoni:
Should be ready for review :)

darosior:
We again have CI timeouts, there is something wrong. Could you add a commit addressing https://github.com/revault/revaultd/issues/186 to this PR so we can inspect what's going on? (should be trivial, add pytest-timeout to README, add a TIMEOUT to the ci script, make sure it does not clash with our own TIMEOUT for the wait_for). I think a timeout of 30min per test is more than enough and will avoid these " Cancelled after 195m" CI runs.

(will review soon :p)

darosior (commented):
I really don't like having the join handle be an `Option<>`. But also, we don't need to join() the thread before responding to the `stop` command, and we probably shouldn't. All that `stop` should do is stopping the JSONRPC server.

Therefore, i think that cleaning up after the jsonrpc_loop in main is better, and having 2 lines of unsafe is preferable over unwraping an option all over the codebase.
```diff
diff --git a/src/daemon/control.rs b/src/daemon/control.rs
index 214d3dc..faf45c4 100644
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -678,7 +678,7 @@ pub fn announce_spend_transaction(
 pub struct RpcUtils {
     pub revaultd: Arc<RwLock<RevaultD>>,
     pub bitcoind_tx: Sender<BitcoindMessageOut>,
-    pub bitcoind_thread: Arc<RwLock<Option<JoinHandle<()>>>>,
+    pub bitcoind_thread: Arc<RwLock<JoinHandle<()>>>,
     pub sigfetcher_tx: Sender<SigFetcherMessageOut>,
-    pub sigfetcher_thread: Arc<RwLock<Option<JoinHandle<()>>>>,
+    pub sigfetcher_thread: Arc<RwLock<JoinHandle<()>>>,
 }
diff --git a/src/daemon/jsonrpc/api.rs b/src/daemon/jsonrpc/api.rs
index d8342f1..d499896 100644
--- a/src/daemon/jsonrpc/api.rs
+++ b/src/daemon/jsonrpc/api.rs
@@ -271,33 +271,7 @@ impl RpcApi for RpcImpl {
             .sigfetcher_tx
             .send(SigFetcherMessageOut::Shutdown)
             .map_err(|e| internal_error!(e))?;
-        assume_ok!(
-            meta.rpc_utils
-                .bitcoind_thread
-                .write()
-                .unwrap()
-                .take()
-                .unwrap()
-                .join(),
-            "Joining bitcoind thread"
-        );
-        assume_ok!(
-            meta.rpc_utils
-                .sigfetcher_thread
-                .write()
-                .unwrap()
-                .take()
-                .unwrap()
-                .join(),
-            "Joining sigfetcher thread"
-        );
-
-        log::info!("Joined");
 
-        // We are always logging to stdout, should it be then piped to the log file (if daemon) or
-        // not. So just make sure that all messages made it to the other end of the connection before
-        // terminating the process.
-        assume_ok!(io::stdout().flush(), "Flushing stdout");
         meta.shutdown();
 
         Ok(())
diff --git a/src/daemon/main.rs b/src/daemon/main.rs
index 399b086..5975cfd 100644
--- a/src/daemon/main.rs
+++ b/src/daemon/main.rs
@@ -22,7 +22,7 @@ use revault_net::sodiumoxide;
 use revault_tx::bitcoin::hashes::hex::ToHex;
 
 use std::{
-    env, panic,
+    env, io::{self, Write}, panic,
     path::PathBuf,
     process,
     sync::{mpsc, Arc, RwLock},
@@ -101,17 +101,34 @@ fn daemon_main(mut revaultd: RevaultD) {
     );
 
     // Handle RPC commands until we die.
+    let bitcoind_thread = Arc::new(RwLock::new(bitcoind_thread));
+    let sigfetcher_thread = Arc::new(RwLock::new(sigfetcher_thread));
     let rpc_utils = RpcUtils {
         revaultd,
         bitcoind_tx,
-        bitcoind_thread: Arc::new(RwLock::new(Some(bitcoind_thread))),
+        bitcoind_thread: bitcoind_thread.clone(),
         sigfetcher_tx,
-        sigfetcher_thread: Arc::new(RwLock::new(Some(sigfetcher_thread))),
+        sigfetcher_thread: bitcoind_thread.clone(),
     };
     assume_ok!(
         rpcserver_loop(socket, user_role, rpc_utils),
         "Error in the main loop"
     );
+
+    // If the RPC server loop stops, we've been told to shutdown!
+    let bitcoind_thread = unsafe {
+        Arc::into_raw(bitcoind_thread).read().into_inner()
+    };
+    let sigfetcher_thread = unsafe {
+        Arc::into_raw(sigfetcher_thread).read().into_inner()
+    };
+    bitcoind_thread.unwrap().join().expect("Joining bitcoind thread");
+    sigfetcher_thread.unwrap().join().expect("Joining sigfetcher thread");
+
+    // We are always logging to stdout, should it be then piped to the log file (if daemon) or
+    // not. So just make sure that all messages made it to the other end of the connection before
+    // terminating the process.
+    assume_ok!(io::stdout().flush(), "Flushing stdout");
 }
 
 // This creates the log file automagically if it doesn't exist, and logs on stdout
```

darosior:
Wow. It's green :no_mouth:

darosior:
Our handling of shutdown on Unix was misguided. Here is a patch special-casing it as synchronous (and moves the check after the actual handler of a read event, as a single event can be both Read and Read Closed !)

```diff
diff --git a/src/daemon/jsonrpc/server.rs b/src/daemon/jsonrpc/server.rs
index a8b704e..0909fd1 100644
--- a/src/daemon/jsonrpc/server.rs
+++ b/src/daemon/jsonrpc/server.rs
@@ -195,19 +195,23 @@ fn read_handle_request(
                 let t_meta = metadata.clone();
                 let t_queue = resp_queue.clone();
 
-                // If there are too many threads spawned, wait for the oldest one to complete.
-                // FIXME: we can be smarter than that..
-                if handler_threads.len() >= MAX_HANDLER_THREADS {
-                    handler_threads
-                        .pop_front()
-                        .expect("Just checked the length")
-                        .join()
-                        .unwrap();
-                }
+                if m.method.as_str() == "stop" {
+                    handle_single_request(t_io_handler, t_meta, t_queue, m);
+                } else {
+                    // If there are too many threads spawned, wait for the oldest one to complete.
+                    // FIXME: we can be smarter than that..
+                    if handler_threads.len() >= MAX_HANDLER_THREADS {
+                        handler_threads
+                            .pop_front()
+                            .expect("Just checked the length")
+                            .join()
+                            .unwrap();
+                    }
 
-                handler_threads.push_back(thread::spawn(move || {
-                    handle_single_request(t_io_handler, t_meta, t_queue, m)
-                }));
+                    handler_threads.push_back(thread::spawn(move || {
+                        handle_single_request(t_io_handler, t_meta, t_queue, m)
+                    }));
+                }
             }
             // Parsing error? Assume it's a message we'll be able to read later.
             Err(e) => {
@@ -252,7 +256,8 @@ fn mio_loop(
     let mut read_cache_map: HashMap<Token, Vec<u8>> = HashMap::with_capacity(8);
     let jsonrpc_io = Arc::from(RwLock::from(jsonrpc_io));
     // Handle to thread currently handling commands we were sent.
-    let mut handler_threads = VecDeque::with_capacity(MAX_HANDLER_THREADS);
+    let mut handler_threads: VecDeque<std::thread::JoinHandle<_>> =
+        VecDeque::with_capacity(MAX_HANDLER_THREADS);
 
     poller
         .registry()
@@ -300,20 +305,6 @@ fn mio_loop(
                     }
                 }
             } else if connections_map.contains_key(&event.token()) {
-                // TODO: determine if it shoudl include event.is_write_closed()
-                if event.is_read_closed() || event.is_error() {
-                    log::trace!("Dropping connection for {:?}", event.token());
-                    connections_map.remove(&event.token());
-
-                    // If this was the last connection alive and we are shutting down,
-                    // actually shut down.
-                    if metadata.is_shutdown() && connections_map.is_empty() {
-                        return Ok(());
-                    }
-
-                    continue;
-                }
-
                 // Under normal circumstances we are always interested in both
                 // Writable (do we got something for them from the resp_queue?)
                 // and Readable (do they have something for us?) events
@@ -327,7 +318,6 @@ fn mio_loop(
                 )?;
 
                 if event.is_readable() {
-                    log::trace!("Readable event for {:?}", event.token());
                     let read_cache = assume_some!(
                         read_cache_map.get_mut(&event.token()),
                         "Entry is always set when connection_map's entry is"
@@ -374,6 +364,20 @@ fn mio_loop(
                         }
                     }
                 }
+
+                if event.is_read_closed() || event.is_error() {
+                    log::trace!("Dropping connection for {:?}", event.token());
+                    connections_map.remove(&event.token());
+
+                    // If this was the last connection alive and we are shutting down,
+                    // actually shut down.
+                    if metadata.is_shutdown() && connections_map.is_empty() {
+                        while let Some(t) = handler_threads.pop_front() {
+                            t.join().unwrap();
+                        }
+                        return Ok(());
+                    }
+                }
             }
         }
     }
@@ -686,7 +690,6 @@ mod tests {
         let msg = String::from(r#"{"jsonrpc": "2.0", "id": 0, "method": "stop", "params": []}"#);
         sock.write(msg.as_bytes()).unwrap();
         sock.flush().unwrap();
-        thread::sleep(Duration::from_secs(1));
         drop(sock);
         server_loop_thread.join().unwrap();
```

darosior:
Applied the above patch in place of the flush fix, 764a796

darosior:
Applied my other patch too..

src/daemon/jsonrpc/api.rs
```diff
@@ -364,43 +452,356 @@ impl RpcApi for RpcImpl {
...
+                    "Invalid signature in Unvault transaction: '{}'",
+                    e
+                ))
+                .to_string(),
+            ));
+        }
+
+        // Sanity checks passed. Store it then share it.
+        // FIXME: edgy edge case: don't crash here, rather return an error if
+        // deposit tx was reorged out in between now and the above status check.
```
darosior, line 752: Looks like you fixed this FIXME :)

src/daemon/jsonrpc/api.rs
```diff
@@ -364,43 +452,356 @@ impl RpcApi for RpcImpl {
...
+        let unvault_descriptor = revaultd
+            .unvault_descriptor
+            .derive(vault.derivation_index, &revaultd.secp_ctx);
+        let cpfp_descriptor = revaultd
+            .cpfp_descriptor
+            .derive(vault.derivation_index, &revaultd.secp_ctx);
+
+        let unvault_tx =
+            UnvaultTransaction::new(deposit_txin, &unvault_descriptor, &cpfp_descriptor, 0)
+                .map_err(|e| internal_error!(e))?;
```
darosior, line 662: Unrelated but since you are touching this line already, can you:
```diff
diff --git a/src/daemon/jsonrpc/api.rs b/src/daemon/jsonrpc/api.rs
index 69456c4..8434ed1 100644
--- a/src/daemon/jsonrpc/api.rs
+++ b/src/daemon/jsonrpc/api.rs
@@ -658,7 +658,7 @@ impl RpcApi for RpcImpl {
             .derive(vault.derivation_index, &revaultd.secp_ctx);
 
         let unvault_tx =
-            UnvaultTransaction::new(deposit_txin, &unvault_descriptor, &cpfp_descriptor, 0)
+            UnvaultTransaction::new(deposit_txin, &unvault_descriptor, &cpfp_descriptor, revaultd.lock_time)
                 .map_err(|e| internal_error!(e))?;
 
         Ok(json!({
```

src/daemon/jsonrpc/api.rs
```diff
@@ -364,43 +452,356 @@ impl RpcApi for RpcImpl {
...
-            return Err(JsonRpcError::invalid_params(err_str));
+            return Err(JsonRpcError::invalid_params(format!(
+                "Invalid signature in Unvault Emergency transaction: {}",
+                e
+            )));
+        }
+
+        // Ok, signatures look legit. Add them to the PSBTs in database.
+        // FIXME: edgy edge case: don't crash here, rather return an error if
+        // deposit tx was reorged out in between now and the above status check.
```
darosior, line 571: Same, you fixed it ! :)

src/daemon/jsonrpc/api.rs
```diff
@@ -575,38 +1084,29 @@ impl RpcApi for RpcImpl {
...
-            "Sending 'delspendtx' to main thread"
-        );
+        let db_path = meta.rpc_utils.revaultd.read().unwrap().db_file();
 
-        assume_ok!(
-            response_rx.recv(),
-            "Receiving 'delspendtx' result from main thread"
-        )
-        .map_err(|e| JsonRpcError::invalid_params(e.to_string()))?;
+        db_delete_spend(&db_path, &spend_txid)
```
darosior, line 1071: Super nit: no need for the blank line (L1088)?

src/daemon/jsonrpc/api.rs
```diff
@@ -616,18 +1116,108 @@ impl RpcApi for RpcImpl {
...
+        let db_path = revaultd.db_file();
+
+        // Get the Spend they reference from DB
+        let mut spend_tx =
+            match db_spend_transaction(&db_path, &spend_txid).map_err(|e| internal_error!(e))? {
+                Some(tx) => tx,
+                None => {
+                    return Err(JsonRpcError::invalid_params(
+                        RpcControlError::UnknownSpend.to_string(),
+                    ));
```
darosior, line 1111: Can you  get rid the RpcControlError ? It has no reason to be anymore.

src/daemon/jsonrpc/server.rs
```diff
@@ -87,13 +87,15 @@ fn read_bytes_from_stream(stream: &mut dyn io::Read) -> Result<Option<Vec<u8>>,
                         }
                         return Ok(Some(trimmed(buf, total_read)));
                     }
-                    io::ErrorKind::Interrupted
-                    | io::ErrorKind::ConnectionReset
+                    io::ErrorKind::Interrupted => {
+                        // Interrupted? Let's try again
```
darosior, line 91: How rude!

darosior (commented):
Looks really good :tada:

Awesome work, i think it's a really nice refactoring and the more it goes the more i think it was the right decision to do it that way: we'll be able to isolate more small routines in `src/daemon/control` which will allow for more thorough unit testing.

As usual, a few nits! Continuing to test this by rebasing my protocol refactorings on top of it.

src/daemon/control.rs
```diff
@@ -135,7 +122,7 @@ fn bitcoind_wallet_tx(
 
 // Tell bitcoind to broadcast the Unvault transactions of all these vaults.
 // The vaults must be active for the Unvault to be finalizable.
-fn bitcoind_broadcast_unvaults(
+pub fn bitcoind_broadcast_unvaults(
```
darosior, line 125: nit: can you for the functions you make public:
1. Make the existing comment doc comments
2. Make sure they are still valid
3. If there is no comment, add a *meaningful* (not necessarily long) doc comment for it?

darosior (commented):
Another thing wrt the doc. We still have a lot to clean up in `control`!

src/daemon/jsonrpc/api.rs
```diff
@@ -221,31 +244,68 @@ macro_rules! parse_vault_status {
     };
 }
 
+macro_rules! internal_error {
+    ($error: expr) => {
+        JsonRpcError {
+            code: jsonrpc_core::types::error::ErrorCode::InternalError,
+            message: format!("Internal error: {:?}", $error),
```
darosior, line 251: nit: i think it's awkward for the user to receive, eg, "Internal error: cosigning server error". Can we keep the placeholder code until we have some dedicated ones but drop the `Internal error` prefix?

darosior (commented):
Another small nit i forgot yesterday

src/daemon/control.rs
```diff
@@ -37,6 +37,7 @@ use revault_tx::{
     },
 };
 
+use jsonrpc_core::types::error::Error as JsonRpcError;
```
darosior, line 40: i'm sorry if i induced you in error, we want `JsonRpcError`s in the `jsonrpc` module but it shoudl absolutely not leak in `control` (that's the point of separating it in the first place!).

Can we have more specialized enums here, pretty much what i tried to do in #196 with the `CommunicationError` and friends ?

src/daemon/control.rs
```diff
@@ -505,7 +513,7 @@ pub fn check_spend_signatures(
             let sig = psbtin
                 .partial_sigs
                 .get(&pubkey)
-                .ok_or_else(|| SigError::MissingSignature(pubkey))?;
+                .ok_or(SigError::MissingSignature(pubkey))?;
```
darosior, line 516: hey, that's clippy who said that? I think it's adding a permanent allocation for nothing? (just curious, and surely wrong)

src/daemon/jsonrpc/api.rs
```diff
@@ -254,6 +254,21 @@ macro_rules! internal_error {
     };
 }
 
+macro_rules! unknown_outpoint {
+    ($outpoint: expr) => {
+        JsonRpcError::invalid_params(format!("No vault at '{}'", $outpoint))
+    };
+}
```
darosior, line 261: :+1:

src/daemon/jsonrpc/api.rs
```diff
@@ -385,19 +399,21 @@ impl RpcApi for RpcImpl {
...
+                JsonRpcError::invalid_params(format!(
                     "'{}' does not refer to a known and confirmed vault",
                     &outpoint,
-                ))),
-                _ => Ok(vault),
-            },
-        }?;
+                ))
+            })?;
+
```
darosior, line 410: nit: gratuitous space?

darosior (commented):
Please, also remove the data structure specific to Rpc from `threadmessages`, that's not their place anymore.. Maybe put them in `control` with a smaller `RpcControlError` which we end up needing anyways for the routines there (https://github.com/revault/revaultd/pull/133#discussion_r619409886).

darosior:
Otherwise looking much better, almost there!

src/daemon/control.rs
danielabrozzoni: clippy says:
```
warning: unnecessary closure used to substitute value for `Option::None`
   --> src/daemon/control.rs:513:23
    |
513 |               let sig = psbtin
    |  _______________________^
514 | |                 .partial_sigs
515 | |                 .get(&pubkey)
516 | |                 .ok_or_else(|| SigError::MissingSignature(pubkey))?;
    | |__________________________________________________________________^
    |
    = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
help: Use `ok_or` instead
    |
513 |             let sig = psbtin
514 |                 .partial_sigs
515 |                 .get(&pubkey).ok_or(SigError::MissingSignature(pubkey))?;
```

I think the point is that, yes, you're adding a permanent allocation instead of re-evaluating a costant every time?

darosior: Yes, makes sense

darosior:
ACK d631f76 :tada:

A few nits (just to be mentioned, will merge without the fixes):
- Use doc comments for the public variants of enums
- Please avoid useless doc comments such as "THis is a presigned tx" for a `PresignedTx` struct :/
- The title of 7beaf1d is now wrong

danielabrozzoni:
You're right 😅 I at least updated the commit title :)

darosior:
ACK 774ced9

darosior:
Sigh, i missed that you kept the `ControlError` and just duplicated it with a `RpcControlError`..... That's ugly

danielabrozzoni:
Agree that it's ugly, I was a bit unsure, but they're treated in two different ways: a ControlError is given back as an internal_error, while a RpcControlError is given back as an invalid params

darosior:
Ok, thanks for the precision. I'll clean it up by getting rid of ControlError and have both RpcParamError and RpcControlError

darosior:
Actually, it depends on how we want to return it in `jsonrpc/api`? Like choosing between the invalid_params and internal_error macros

PR #139, opened 2021-03-30
State tracking for Unvault and Spend transactions

darosior:
This (for once is not based on another PR  and) makes us track the Unvault UTXOs pretty much in the same manner as we already track the deposits. This also features some small drive-by cleanups in the db, revaultd and reorg tests.

We try to not overload `bitcoind` with `listunspent` calls and therefore to track unconfirmed, confirmed and spent utxos in 3 mappings for each descriptor.
The idea for the state tracking is to treat our transaction chain as a tree and to only track the trunk's descriptors on the bitcoind watchonly wallet, then figure out the branches out of it:
```
                         -- Bypass
                     /
deposit utxos  --  Unvault
                     \  Emergency

                           Cancel
                     /        Emergency
unvault utxos     /
                     \ __  Spend
```

For Spend transaction, we use a bulky check similar to the broadcast: every new tip update we check if any new Spend transaction confirmed.

PR #143, opened 2021-03-31
Vault states for signed-by-us presigned transactions

### LINKED ISSUE #105: Did i already sign this ? Let the user know if they need to acknowledge / active this vault
> darosior
20:07
Hmm we should have a "i already signed" state don't you think? Weird to still have to acknowledge funds on homepage after succesful signing

> edouard
09:43
> yes, I can check for each vault not spent if the presigned transactions are signed  but If there are 6371 vaults it would take time to load
> maybe once revocationtxs is done, we update a column in vault table with: revocation_txs_signed: true
and the column has a field is in the resource of the listvaults call

> darosior
09:45
That's basically equivalent to having a new state

> edouard
09:46
yes, do we both agree that a new vault status is too much ?
two other column in the table could do the job:
revocation_txs_signed unvault_tx_signed

> darosior
09:55
Well i feel like the opposite (as usual lol), adding more columns seems overkiller than a new state. But the number of state is bloated. I'm thinking of a way to do it without DB
Let's put an issue for now

darosior:
Fixes #105

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -68,17 +68,19 @@ Get an address to build a deposit transaction.
 | ----- | -------------------- | ------------------------------------------------------------------------------------------------------------ |
 | 0     | `unconfirmed`        | The vault's deposit transaction is less than 6 blocks-deep in the chain                                      |
 | 1     | `funded`             | The vault is initiated by a deposit transaction                                                              |
-| 2     | `secured`            | The vault's emergency transaction is fully signed and shared with the watchtowers                            |
-| 3     | `active`             | The vault's unvault, cancel, and unvault-emergency txs are fully signed and shared with the watchtowers      |
-| 4     | `unvaulting`         | The vault has its unvault tx broadcasted                                                                     |
-| 5     | `unvaulted`          | The vault has its unvault tx confirmed                                                                       |
-| 6     | `cancelling`         | The vault has its cancel tx broadcasted, funds are sent to an other vault                                    |
-| 7     | `cancelled`          | The vault has its cancel tx confirmed, funds are in an other vault                                           |
-| 3 / 6 | `emergency_vaulting` | The vault has its emergency tx broadcasted, funds are sent to the Deep Emergency Vault                       |
-| 4 / 7 | `emergency_vaulted`  | The vault has its emergency tx confirmed, funds are in the Deep Emergency Vault                              |
-| 6     | `spendable`          | The vault has its unvault tx timelock expired and can be spent                                               |
-| 7     | `spending`           | The vault has a spending tx broadcasted                                                                      |
-| 8     | `spent`              | The vault has a spending tx confirmed, the vault is spent                                                    |
+| 2     | `securing`           | We signed and shared the revocation transactions signatures for this vault                                   |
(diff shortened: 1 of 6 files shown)


darosior:
On top of #139 obviously

doc/API.md
```diff
@@ -68,17 +68,19 @@ Get an address to build a deposit transaction.
...
-| 6     | `cancelling`         | The vault has its cancel tx broadcasted, funds are sent to an other vault                                    |
-| 7     | `cancelled`          | The vault has its cancel tx confirmed, funds are in an other vault                                           |
-| 3 / 6 | `emergency_vaulting` | The vault has its emergency tx broadcasted, funds are sent to the Deep Emergency Vault                       |
-| 4 / 7 | `emergency_vaulted`  | The vault has its emergency tx confirmed, funds are in the Deep Emergency Vault                              |
-| 6     | `spendable`          | The vault has its unvault tx timelock expired and can be spent                                               |
-| 7     | `spending`           | The vault has a spending tx broadcasted                                                                      |
-| 8     | `spent`              | The vault has a spending tx confirmed, the vault is spent                                                    |
+| 2     | `securing`           | We signed and shared the revocation transactions signatures for this vault                                   |
```
edouardparis, line 71: maybe we can do the change now for `acknowledging` and `acknowledge` ?

tests/test_misc.py
```diff
@@ -486,6 +496,9 @@ def sign_revocation_txs(stks, deposit):
         unvault_psbt = stk.rpc.getunvaulttx(deposit)["unvault_tx"]
         unvault_psbt = stk.stk_keychain.sign_unvault_psbt(unvault_psbt, child_index)
         stk.rpc.unvaulttx(deposit, unvault_psbt)
+        assert (
+            len(stk.rpc.listvaults(["activating", "active"], [deposit])["vaults"]) == 1
```
edouardparis, line 500: Why not only `activating` ?

doc/API.md
edouardparis: ok ignore comment

tests/test_misc.py
darosior: Because they may have been polling the signature in the background before.

darosior:
Ok, rebased and applied a correction for reorgs status checking
```patch
diff --git a/src/daemon/bitcoind/actions.rs b/src/daemon/bitcoind/actions.rs
index 6d1910a..301cd31 100644
--- a/src/daemon/bitcoind/actions.rs
+++ b/src/daemon/bitcoind/actions.rs
@@ -354,7 +354,11 @@ fn unconfirm_vault(
     // For these states we still have the entry in the deposit cache.
     if matches!(
         vault.status,
-        VaultStatus::Funded | VaultStatus::Secured | VaultStatus::Active
+        VaultStatus::Funded
+            | VaultStatus::Securing
+            | VaultStatus::Secured
+            | VaultStatus::Activating
+            | VaultStatus::Active
     ) {
         deposits_cache
             .get_mut(&vault.deposit_outpoint)
```

edouardparis:
ACK 92b09cd65fe35436e2606cd045fe51a4f1f8e034

edouardparis (approved):
ACK 92b09cd65fe35436e2606cd045fe51a4f1f8e034

PR #144, opened 2021-03-31
Add revault command

danielabrozzoni:
Closes #126

PR #152, opened 2021-04-05
Update revault tx

darosior:
This updates `revault_tx` to the latest version and does some consequent cleanups.

Based on #144

### DIFF
--- a/src/daemon/bitcoind/actions.rs
+++ b/src/daemon/bitcoind/actions.rs
@@ -258,7 +258,7 @@ fn maybe_broadcast_spend_transactions(
 
     for db_spendtx in db_broadcastable_spend_transactions(&db_path)? {
         let mut psbt = db_spendtx.psbt;
-        let txid = psbt.inner_tx().global.unsigned_tx.txid();
+        let txid = psbt.txid();
         log::debug!("Trying to broadcast Spend tx '{}'", &txid);
 
         match psbt.finalize(&revaultd.read().unwrap().secp_ctx) {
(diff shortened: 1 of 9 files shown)


darosior:
Still marked as WIP cause upstream is not merged but should be in a reviewable state.

src/daemon/bitcoind/actions.rs
```diff
@@ -323,6 +324,30 @@ fn mark_confirmed_spends(
...
 }
 
+fn mark_confirmed_cancels(
+    revaultd: &Arc<RwLock<RevaultD>>,
+    bitcoind: &BitcoinD,
+) -> Result<(), BitcoindError> {
+    let db_path = revaultd.read().unwrap().db_file();
+
+    for (db_vault, cancel_tx) in db_canceling_vaults(&db_path)? {
+        let cancel_tx = cancel_tx.into_psbt().extract_tx();
```
danielabrozzoni, line 334: I think `txid()` can be used here too. Can you include the fix in be83f4f1b6d90e92d62ad113dd2c2563b886f623?

darosior: Right, good catch

darosior:
Rebased on top of #144

darosior:
Rebased after upstream merge, this is now ready for review (of course two tests still fail because of #144 )

danielabrozzoni:
utACK 8fbae96a74d751f309447f7b5e1a6360b5a46b91 I read the code multiple times, it seems fine, I didn't test it directly

PR #156, opened 2021-04-07
Bugfixes for dust deposits and `listpresignedtransactions`

darosior:
As experienced during the demo test. Based on #152 

Fixes #148 
Fixes #146

PR #157, opened 2021-04-08
Test derived second stage transactions

### LINKED ISSUE #147: Invalid Cancel and Unvault Emergency signatures for derivation index = 1
I think we have the generation and check of sigs off

darosior:
This adds a test to fix #147 . See commit messages, what we experienced was because `bitcoind` can't understand our Unvault Scripts.

Based on #156

### DIFF
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -17,7 +17,7 @@ use crate::{
         interface::{
             db_cancel_transaction, db_emer_transaction, db_list_spends, db_spend_transaction,
             db_tip, db_unvault_emer_transaction, db_unvault_transaction, db_vault_by_deposit,
-            db_vault_by_unvault_txid, db_vaults, db_vaults_from_spend, db_vaults_min_status
+            db_vault_by_unvault_txid, db_vaults, db_vaults_from_spend, db_vaults_min_status,
         },
         schema::DbVault,
         DatabaseError,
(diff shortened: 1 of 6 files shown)


darosior:
Note there are a few worrying logs there wrt to vanished deposits when canceling, probably due to a logic mistake in #144

src/daemon/control.rs
```diff
@@ -263,19 +263,59 @@ fn presigned_txs_list_from_outpoints(
...
+    // For each presigned transaction, append it as well as its extracted version if it's final.
     let mut tx_list = Vec::with_capacity(db_vaults.len());
     for db_vault in db_vaults {
         let outpoint = db_vault.deposit_outpoint;
 
-        let (_, unvault) = db_unvault_transaction(db_path, db_vault.id)?;
+        let (_, unvault_psbt) = db_unvault_transaction(db_path, db_vault.id)?;
+        let mut finalized_unvault = unvault_psbt.clone();
+        let unvault = VaultPresignedTransaction {
+            transaction: if finalized_unvault.finalize(&revaultd.secp_ctx).is_ok() {
```
danielabrozzoni, line 274: nit: I haven't tested it but I think you can just use map
```
transaction: finalized_unvault.finalize(&revaultd.secp_ctx).map(|f| f.into_psbt().extract_tx())
```

darosior: Nope because finalize modifies the tx in place and does not return it

darosior:
Rebased on master

danielabrozzoni:
utACK 388b99073bfc1cc10dc00b3a66eb1f030246fb4a looks good, didn't test

Issue #35, opened 2020-11-26
Check watchonly wallet at startup


darosior:
So we don't bypass an import in case of a crash or w/e..

darosior:
We'll need bitcoind 0.22 to list imported descritptors, removing from the milestone.

PR #158, opened 2021-04-08
Bulk sharing of presigned transactions signatures at startup

### LINKED ISSUE #96: Startup check that the coordinator got all our sigs
In the `unvaulttx` call (as well as in the `revocationtxs` call for that matter) we *might* crash or have a network connection error just after storing the sig in db, resulting in us having the sig but the others not being able to fetch them.

At startup we could `getsigs` for our sigs to make sure the coordinator got everything already.

darosior:
It's bulky but certainly useful for now. 
Based on #157 

Fixes #96
Adds a TODO for #35 which can't be fixed yet


darosior:
Applied linter and rebased on master

danielabrozzoni:
utACK 59c70f859ee38937ac838194d9c41f84e2104046 reviewed briefly the code, looks good

PR #165, opened 2021-04-09
Unflake functional tests

darosior:

PR #169, opened 2021-04-12
Vault statuses and bitcoind poll fixups

darosior:
Fixes #163
Fixes #161

PR #170, opened 2021-04-12
Test sig fetcher robustness

### LINKED ISSUE #164: Harden test the signature fetching
Activate a vault, then delete its database, restart the daemon and rescan the chain. It should see the deposit, fetch the sigs from the coordinator and mark the vault back as secured then active.

danielabrozzoni:
Based on #165, closes #164

I still need to test it a bit more, as the test sometimes fails as the manager's revaultd dies

### DIFF
--- a/tests/test_framework/revaultd.py
+++ b/tests/test_framework/revaultd.py
@@ -33,18 +33,18 @@ class Revaultd(TailableProc):
 
         # The data is stored in a per-network directory. We need to create it
         # in order to write the Noise private key
-        datadir_with_network = os.path.join(datadir, "regtest")
-        os.makedirs(datadir_with_network, exist_ok=True)
+        self.datadir_with_network = os.path.join(datadir, "regtest")
+        os.makedirs(self.datadir_with_network, exist_ok=True)
 
         bin = os.path.join(
             os.path.dirname(__file__), "..", "..", "target/debug/revaultd"
         )
         self.conf_file = os.path.join(datadir, "config.toml")
         self.cmd_line = [bin, "--conf", f"{self.conf_file}"]
-        socket_path = os.path.join(datadir_with_network, "revaultd_rpc")
+        socket_path = os.path.join(self.datadir_with_network, "revaultd_rpc")
         self.rpc = UnixDomainSocketRpc(socket_path)
 
-        noise_secret_file = os.path.join(datadir_with_network, "noise_secret")
+        noise_secret_file = os.path.join(self.datadir_with_network, "noise_secret")
         with open(noise_secret_file, "wb") as f:
             f.write(noise_priv)
 
--- a/tests/test_framework/utils.py
+++ b/tests/test_framework/utils.py
@@ -32,10 +32,17 @@ LOG_LEVEL = os.getenv("LOG_LEVEL", "trace")
 assert LOG_LEVEL in ["trace", "debug", "info", "warn", "error"]
 
 
-def wait_for(success, timeout=TIMEOUT):
+def wait_for(success, timeout=TIMEOUT, debug_fn=None):
+    """
+        Run success() either until it returns True, or until the timeout is reached.
+        debug_fn is logged at each call to success, it can be useful for debugging
+        when tests fail.
+    """
     start_time = time.time()
     interval = 0.25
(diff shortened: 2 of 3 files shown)


danielabrozzoni:
Failing even after rebasing on top of #169, sometimes manager dies and the log doesn't help at all (they're trimmed). Seems that I have to test in a real setup :/

danielabrozzoni:
Hopefully it's fixed (CI hates me though)

tests/test_framework/revaultd.py
```diff
@@ -33,18 +33,18 @@ def __init__(
 
         # The data is stored in a per-network directory. We need to create it
         # in order to write the Noise private key
-        datadir_with_network = os.path.join(datadir, "regtest")
-        os.makedirs(datadir_with_network, exist_ok=True)
+        self.datadir_with_network = os.path.join(datadir, "regtest")
```
darosior, line 36: bbf3e14d02d35b8d44558fa120be25bf5aba3ef5 title: it's an object attribute, not a class one :)

tests/test_framework/utils.py
```diff
@@ -32,10 +32,17 @@
 assert LOG_LEVEL in ["trace", "debug", "info", "warn", "error"]
 
 
-def wait_for(success, timeout=TIMEOUT):
+def wait_for(success, timeout=TIMEOUT, debug_fn=None):
+    """
+        Run success() either until it returns True, or until the timeout is reached.
+        debug_fn is logged at each call to success, it can be useful for debugging
+        when tests fail.
+    """
```
darosior, line 40: Neat

tests/test_misc.py
```diff
@@ -2101,3 +2103,64 @@ def test_retrieve_vault_status(revault_network, bitcoind):
...
+    revault_network.activate_vault(vault)
+
+    # Stopping revaultd
+    for w in stks + mans:
+        w.stop()
+
+    # Deleting the database
+    for w in stks + mans:
+        datadir_db = os.path.join(w.datadir_with_network, "revaultd.sqlite3")
+        os.remove(datadir_db)
```
darosior, line 2122: nit: Hmm, you don't really need to loop twice?

tests/test_misc.py
```diff
@@ -2101,3 +2103,64 @@ def test_retrieve_vault_status(revault_network, bitcoind):
...
+            ]
+        )
+
+    for w in stks + mans:
+        w.wait_for_log(
+            "Got a new unconfirmed deposit"
+        )
+
+    for w in stks + mans:
+        wait_for(lambda: len(w.rpc.listvaults(["funded"], [])) == 1)
```
darosior, line 2145: This can be in the same loop?

tests/test_misc.py
```diff
@@ -2101,3 +2103,64 @@ def test_retrieve_vault_status(revault_network, bitcoind):
     mans[0].wait_for_secured_vaults([deposit])
 
     # TODO: same dance with all emergency statuses
+
+@pytest.mark.skipif(not POSTGRES_IS_SETUP, reason="Needs Postgres for servers db")
+def test_sigfetcher(revault_network, bitcoind, executor):
+    (stks, mans) = revault_network.deploy(2, 1)
```
darosior, line 2109: Can we do a larger setup, like `7, 3` to make it closer to a realistic one which we want to sanity check?

*(yeah, i know, it does not really make sense but y'know, edgecases, esoteric coordinator, all that)*

darosior (approved):
Looks good to me! A few nits which should be trivial to fix

darosior:
Oo CI is back to life

tests/test_misc.py
danielabrozzoni: Right, but I'm keeping "stopping and deleting" and "starting again" as separate loops, I think it's cleaner

darosior: I don't think so, but i don't really care. Just hearing someone yelling at me INEFFICIENT each time i write an unnecessary loop (which is probably often wrong in Python)

darosior:
ACK 2445d33

darosior:
Hmm CI is dying again but i never had a single die locally (ran the tests hundreds of times this week)

PR #179, opened 2021-04-14
bitcoind: unload any leftover watchonly wallet at startup

### LINKED ISSUE #176: bitcoind watchonly wallet sanity checks

darosior:
We used to have this check and i removed it since it does not seem right
for us to unload wallets on the host's bitcoind. But it's actually fine
if it's the very same one we create (by name) and we'll now also refuse
to start with an inconsistent number of loaded watchonly wallets.

Fixes #176

### DIFF
--- a/src/daemon/bitcoind/interface.rs
+++ b/src/daemon/bitcoind/interface.rs
@@ -230,6 +230,25 @@ impl BitcoinD {
         )))
     }
 
+    pub fn unloadwallet(&self, wallet_path: String) -> Result<(), BitcoindError> {
+        let res = self.make_node_request("unloadwallet", &params!(Json::String(wallet_path),))?;
+
+        let warning = res
+            .get("warning")
+            .map(|w| w.as_str())
+            .flatten()
+            .ok_or_else(|| {
+                BitcoindError::Custom(
+                    "No or invalid 'warning' in 'unloadwallet' result".to_string(),
+                )
+            })?;
+        if warning.len() > 0 {
+            Err(BitcoindError::Custom(warning.to_string()))
+        } else {
+            Ok(())
+        }
+    }
+
     /// Constructs an `addr()` descriptor out of an address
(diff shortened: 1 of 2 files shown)


darosior:
Fuck i removed my interpretor hook for rustfmt :sweat_smile: Here you go

src/daemon/bitcoind/interface.rs
```diff
@@ -230,6 +230,16 @@ impl BitcoinD {
         )))
     }
 
+    pub fn unloadwallet(&self, wallet_path: String) -> Result<(), BitcoindError> {
+        let res = self.make_node_request("unloadwallet", &params!(Json::String(wallet_path),))?;
+
+        if let Some(warning) = res.get("warning").map(|s| s.to_string()) {
```
danielabrozzoni, line 236: You have to check if warning is different from `""`. From the docs:
```
bitcoin-cli help unloadwallet 
unloadwallet ( "wallet_name" load_on_startup )
Unloads the wallet referenced by the request endpoint otherwise unloads the wallet specified in the argument.
Specifying the wallet name on a wallet endpoint is invalid.
Arguments:
1. wallet_name        (string, optional, default=the wallet name from the RPC endpoint) The name of the wallet to unload. Must be provided in the RPC endpoint or this parameter (but not both).
2. load_on_startup    (boolean, optional, default=null) Save wallet name to persistent settings and load on startup. True to add wallet to startup list, false to remove, null to leave unchanged.

Result:
{                       (json object)
  "warning" : "str"     (string) Warning message if wallet was not unloaded cleanly.
}
```
So `warning` is always there, and if everything is fine it's equal to `""`

darosior: Oh, right, thanks

darosior: Fixed

danielabrozzoni (approved):
ACK 1bbfb87309d8e021f205b10701974ea17ad43df9

PR #181, opened 2021-04-14
Fix human readable CLI output

### LINKED ISSUE #162: Only return the `result` on human readable CLI
We have unnecessary info right now:
```
{
  "id": "revault-cli-3252059",
  "jsonrpc": "2.0",
  "result": {
    "vaults": [
    ```

### PR DESCRIPTION
**Pull request by danielabrozzoni**
Returning only the content of `result` if success, otherwise returning only `error`

 Fixes #162

### DIFF
--- a/src/cli/main.rs
+++ b/src/cli/main.rs
@@ -161,7 +161,17 @@ fn main() {
                     if raw {
                         print!("{}", response);
                     } else {
-                        print!("{:#}\n", response);
+                        if let Some(r) = response.get("result") {
+                            println!("{:#}", serde_json::json!({ "result": r }));
+                        } else if let Some(e) = response.get("error") {
+                            println!("{:#}", serde_json::json!({ "error": e }));
+                        } else {
+                            log::warn!(
+                                "revaultd response doesn't contain result or error: '{}'",
+                                response
+                            );
+                            println!("{:#}", response);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/cli/main.rs [hunk 1]
```diff
@@ -161,7 +161,14 @@ fn main() {
                     if raw {
                         print!("{}", response);
                     } else {
-                        print!("{:#}\n", response);
+                        if let Some(r) = response.get("result") {
+                            // If the response is null, don't print anything
+                            if Json::Null != *r {
+                                println!("{:#}", r);
+                            }
```
darosior, line 168: Shouldn't we be printing `null` ? Anyways it's a violation of the JSONRPC2 spec to not return an object as result iirc

darosior (approved):
ACK aaddcf2f86b5c1acc04077c7f47983127d988453

Bikeshedding on the `null`, but meh

src/cli/main.rs
danielabrozzoni: If you prefer we can return `null`, I'm not really sure what's the convention here

darosior: Yeah, let's do that. Don't know what's best either but when in doubt Explicit Is Always Better :tm:

darosior:
ACK 181

PR #184, opened 2021-04-16
bitcoind: only forget deposit that are really gone

darosior:
Still trying to reproduce the panic without the patch, but this should at least fix it for master


darosior:
ACK 799f33b

PR #189, opened 2021-04-22
ci: Add pytest timeout

danielabrozzoni:


contrib/ci-functional-tests.sh
```diff
@@ -27,4 +27,4 @@ cd tests/servers/cosignerd && cargo build && cd "$REPO_ROOT"
 python3 -m venv venv
 . venv/bin/activate
 pip install -r tests/requirements.txt
-VERBOSE=1 LOG_LEVEL=debug TIMEOUT=120 TEST_DEBUG=1 POSTGRES_USER="test" POSTGRES_PASS="test" pytest -n3 -vvv --log-cli-level=DEBUG tests/
+VERBOSE=1 LOG_LEVEL=debug TIMEOUT=120 TEST_DEBUG=1 POSTGRES_USER="test" POSTGRES_PASS="test" pytest -n3 -vvv --log-cli-level=DEBUG --timeout=600 tests/
```
darosior, line 30: Ten minutes is not enough for the CI, we have pretty slow tests. `1800` seems reasonable: if it gets over this timeout, we are sure it's something with our tests not the CI being atrociously slow

tests/test_framework/utils.py
```diff
@@ -457,6 +457,8 @@ def wait_for_logs(self, regexs, timeout=TIMEOUT):
 
             with self.logs_cond:
                 if pos >= len(self.logs):
+                    if not self.running:
+                        raise ValueError("Process died while waiting for logs")
```
darosior, line 461: But, what if the process dies without printing the logs you were looking for ?

contrib/ci-functional-tests.sh
danielabrozzoni: Just updated, I hoped that 10 mins were enough :(

tests/test_framework/utils.py
danielabrozzoni: This is checking if the process is dead when it's done reading the logs. So if I don't find the logs I was looking for I'm going to read all of them till the last one, check if the process is dead, if it is raise an error

danielabrozzoni: The point is that if you have 10 lines of logs you have to read all of them before saying "Process died while waiting for logs".
An example: we're looking for "Stopping revaultd", which is located in the logs at line 8
Python is slow at processing logs, when it's processing line 5 it notices that the process is dead and raises an error (happens to me every time in local)
This is wrong, because the log is there. Python should process every log and just then raise an error if the daemon is dead

darosior: Oh, right. Makes sense. But iirc i had a patch for this which actually made this check an `elif` of:
```python
            if timeout is not None and time.time() > start_time + timeout:
                print("Time-out: can't find {} in logs".format(exs))
                for r in exs:
                    if self.is_in_log(r):
                        print("({} was previously in logs!)".format(r))
                 raise TimeoutError('Unable to find "{}" in logs.'.format(exs))
```
in #133 ? Maybe not. Anyways this or checking just above looks good, let's see what CI says with timeouts.

danielabrozzoni: I cherry-picked this commit from #133 as I expect to merge this one before and this is not really related to the json rpc thread anyway 
There is another commit in #133 that helps with tests: https://github.com/revault/revaultd/pull/133/commits/291c60d0c0d539782f38ff9cf93020329f029800
Can't find anything else

darosior:
ACK dc60abe

The timeout found something!
> Error:   root:utils.py:414 revaultd-man-1: [04-22][10:29:23][revaultd::bitcoind::actions][ERROR] Error in bitcoind poller thread: 'Custom("Deposit transaction isn\'t confirmed!")'

darosior:
What the fuck for this error though

darosior:
https://github.com/revault/revaultd/issues/190

darosior:
Aaaarg wtf this rebase too :smile:

darosior:
Rebased on master

darosior:
ACK c152273

PR #195, opened 2021-04-22
Testing windows ci :)

danielabrozzoni:


src/daemon/jsonrpc/server.rs
```diff
@@ -87,13 +87,15 @@ fn read_bytes_from_stream(stream: &mut dyn io::Read) -> Result<Option<Vec<u8>>,
...
+                        // Interrupted? Let's try again
+                        continue;
+                    }
+                    io::ErrorKind::ConnectionReset
                     | io::ErrorKind::ConnectionAborted
                     | io::ErrorKind::BrokenPipe => {
-                        // Try again on interruption or disconnection. In the latter case we'll
-                        // remove the stream anyways.
-                        continue;
+                        // Something strange happened, but it's not fatal
```
darosior, line 97: It's not "something strange", they just closed on their end :)

danielabrozzoni: Yeah the previous comment made me wonder if something different had to happen on disconnection ("we'll remove the stream anyways"?), but you're right

danielabrozzoni:
Closing, I needed it just for testing anyway

PR #198, opened 2021-04-23
doc: add gethistory to API.md

edouardparis:
for example:
- 2021-01-01 one vault V1 deposited
- 2021-01-02 one vault V2 deposited
- 2021-01-03 one vault V3 deposited
- 2021-01-04 spend transaction S1 spent V1 and V2

We should have the response of `gethistory [deposit, cancel, spend] 123412342352 20`
```json
{
  "events": [
      {
        "blockheight": 112236,
        "received_at": 123412342352, 
        "hex": "hex S1 tx",
        "kind": "spend",
        "vault_outpoints": ["v1::outpoint", "v2::outpoint"]
      },
      {
        "blockheight": 112234,
        "received_at": 123412342352, 
        "hex": "hex v3 tx",
        "kind": "deposit"
      },
      {
        "blockheight": 112233,
        "received_at": 123412342352, 
        "hex": "hex v2 tx",
        "kind": "deposit"
      },
      {
        "blockheight": 112232,
        "received_at": 123412342352, 
        "hex": "hex v1 tx",
        "kind": "deposit"
      }
  ]
}
```

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -16,10 +16,11 @@ Note that all addresses are bech32-encoded *version 0* native Segwit `scriptPubK
 | [`listvaults`](#listvaults)                                 | Display a paginated list of vaults                   |
 | [`revocationtxs`](#revocationtxs)                           | Give back the revocation transactions signed         |
 | [`unvaulttx`](#unvaulttx)                                   | Give back the unvault transaction signed             |
-| [`updatespendtx`](#updatespendtx)                            | Store or update the stored Spend transaction         |
+| [`updatespendtx`](#updatespendtx)                           | Store or update the stored Spend transaction         |
 | [`delspendtx`](#delspendtx)                                 | Delete a stored Spend transaction                    |
 | [`setspendtx`](#setspendtx)                                 | Announce and broadcast this Spend transaction        |
 | [`listspendtxs`](#listspendtxs)                             | List all stored Spend transactions                   |
+| [`gethistory`](#gethistory)                                 | Retrieve history of funds                            |
 
 
 
@@ -380,6 +381,35 @@ You'll have to manually fetch the vaults statuses if you want to know, for examp
 None; the `result` field will be set to the empty object `{}`. Any value should be
 disregarded for forward compatibility.
 
+### `gethistory`
+
+`gethistory` retrieves a paginated list of accounting events.
+
+Aiming at giving an accounting point of view, the amounts returned by this call are the total
+of inflows and outflows net of any change amount (that is technically a transaction output, but not a cash outflow).
+
+#### Request
+
+| Field    | Type         | Description                                                          |
+| -------- | ------------ | -------------------------------------------------------------------- |
+| `cursor` | int          | Timestamp of the date to retrieve events after                       |
+| `limit`  | int          | Number of events to retrieve                                         |
+| `kind`   | string array | Type of the events to retrieve, can be `deposit`, `cancel`, `spend`  |
+
+#### Response
+
+| Field          | Type   | Description                                |
+| -------------- | ------ | ------------------------------------------ |
+| `events`       | array  | Array of [Event resource](#event-resource) |
+
+##### Event Resource
+
+| Field    | Type   | Description                                                                              |
+| -------- | ------ | ---------------------------------------------------------------------------------------- |
+| `kind`   | string | Type of the event, can be `deposit`, `cancel`, `spend`                                   |
+| `date`   | int    | Timestamp of the event                                                                   |
+| `amount` | int    | Absolute amount in satoshis that is entering or exiting the wallet                       |
+| `fee`    | int    | Fee of the event transaction                                                             |
 
 ## User flows
 


darosior:
Instead of a raw transaction, we could give you the amounts instead? It seems more intuitive for a financial history. Why should we return the blockheight for ? I don't think it'd useful to your panel and i'm not sure if it would be to other consumers of the API.

All in all it seems to me that it's pretty similar to the others calls in its current form rather than being "accountancy-centric". However i like the differentiation to the other commands that you can have single vault be present twice here.

edouardparis:
Agree to remove blockheight.
the Gui will need the outputs address and the amounts if the hex is not in the response for a spend transaction:
```json
{
        "received_at": 123412342352, 
        "recipients": {
            "address1":  4234234,
            ...
        },
        "kind": "spend",
        "vault_outpoints": ["v1::outpoint", "v2::outpoint"]
      },
```

darosior:
How about having:
- The event kind
- The event timestamp
- The outgoing amount
- The fees (with `fees == outgoing amount` if this is the `kind` is `cancel`)

edouardparis:
That is exactly what is needed for the GUI :+1:
I thought some parts would be done by the GUI, like the fees calcul, but If the daemon can do all the abstraction to get this result, It is even better for me

darosior:
Hmm i think from an API perspective it makes sense to have it in the daemon, as other clients than the GUI may be interested in this (like an accountancy software). Exposing information that is already available through other calls and having all the clients of our API do the same job seems bad.

doc/API.md
```diff
@@ -368,6 +369,30 @@ disregarded for forward compatibility.
...
+| `events`       | array  | Array of [Event resource](#event-resource) |
+
+##### Event Resource
+
+| Field    | Type   | Description                                                                              |
+| -------- | ------ | ---------------------------------------------------------------------------------------- |
+| `kind`   | string | Type of the event, can be `deposit`, `cancel`, `spend`                                   |
+| `date`   | int    | Timestamp of the event                                                                   |
+| `amount` | int    | Outgoing or incoming Amount of the event transaction, `null` if the event is a `cancel`  |
+| `fee`    | int    | Fee of the event transaction, `null` if the event is a `deposit`                         |
```
darosior, line 395: I think `amount` should always be there, as a total. And the `fee` be an optional part of it. For instance:
- For a deposit of 1M sat, `amount: 1_000_000`, `fee: null`
- For a cancel of 1M sat, `amount: `1_000_000`, `fee: 1000`
- For a spend of 1M sat, `amount: 1_000_000`, `fee: 2000`

Maybe for the spend it's awkward to have the `amount` not be the "spent amount" but i believe it makes sense to put the entire amount the previously-deposited vault.

darosior (commented):
So the CPFP outputs value would count as `fee` ? I think it makes sense.

doc/API.md
edouardparis: No I think it is confusing, notably for the cancel: what does amount means ? only fees are outgoing 

For the spend: do you count the spend output in the amount and the history has an other row with the change output amount as deposit ?

edouardparis: It is difficult for the user to understand the move in wallet, maybe we can think of something like:
| kind    | fee  | amount   |
| ------  | ---- | -------  |
| cancel  | 1000 |          |
| spend   | 1000 | -1999000 |
| deposit |      | +2000000 |

darosior: > what does amount means?

`amount` is the value of a vault.

From an accountancy point of view it seemed odd to me that you would have a deposit then a spend event that would not be of the same amount. I also dislike having nullable fields so reducing their number would be nice.

> [...] what about the change value?

Arg. Good point. Ok, i think your solution is the only thing left :)

edouardparis: or we can seperate amount by `moved amount`, `incoming_amount` and `outgoing amount` and no null value only 0
| kind    | fee  | amount moved | incoming amount | outgoing_amount |
| ------  | ---- | ------------ | --------------  | --------------- |
| cancel  | 1000 | 100000       | 0               | 0               |
| spend   | 1000 | 2000000      | 0               | 1989000         |
| deposit |    0 | 2000000      | 2000000         | 0               |

In this case, the user can deduce that a change output was created by the spend: `moved amount > outgoing_amount + fee`

edouardparis: Or sorry maybe moved amount can only be amount already in vault and we can call also incoming amount the change output amount:
in this case we have:
| kind    | fee  | amount moved | incoming amount | outgoing_amount |
| ------  | ---- | ------------ | --------------  | --------------- |
| cancel  | 1000 | 100000       | 99000           | 0               |
| spend   | 1000 | 2000000      | 10000           | 1989000         |
| deposit |    0 | 0            | 2000000         | 0               |

darosior: I don't think that the accountancy tab should have a concept of what really happens onchain (movements). It should rather abstracts it to "we sent X-Y" rather than "we moved X and got Y back".

edouardparis: I think the last level of abstraction can be done by the client, I think some other usecase will use this method, like overall statistics of volume of funds moving.}
Maybe one day we will have a new feature "consolidation" with a new line:
| kind    | fee  | amount moved | incoming amount | outgoing_amount |
| ------  | ---- | ------------ | --------------  | --------------- |
| cancel  | 1000 | 100000       | 99000           | 0               |
| spend   | 1000 | 2000000      | 10000           | 1989000         |
| deposit |    0 | 0            | 2000000         | 0               |
| consolidation | 3000 | 250000  | 247000| 0 |

danielabrozzoni: I think it's clearer if you just use `amount` with positive and negative values instead of incoming and outcoming, as you said:

kind | fee | amount
-- | -- | --
cancel | 1000 |  -0
spend | 1000 | -1999000
deposit | 0 | +2000000

Substituting the null value with "-0" (or "+0"? I don't know)
And maybe putting a negative number under "fee"? Like:

kind | fee | amount
-- | -- | --
cancel | -1000 |  -0
spend | -1000 | -1999000
deposit | -0 | +2000000

darosior: I like the negative idea, but not for the fees. Conceptually it's weird to me to have a "negative fee" (you are paid for using it?)

danielabrozzoni:
Yeah agree, that's why I wasn't really sure. Let's keep the fee ""unsigned"" (just "1000", without + or -)

doc/API.md
edouardparis: OK https://github.com/revault/revaultd/pull/198#issuecomment-831081469
let's go for unsigned value, we can deduce from the `kind` if it is incoming or leaving

I wanted to keep the two columns to have information about change output for a spend, I wanted that GUI display both columns but aggregate like you do in the column `amount` and when a user click on it, he gets the detail with amount of fund moved, change amount etc

danielabrozzoni: I would keep the CLI clean, even if it's just for us, and avoid having both the "amount" and "incoming/outcoming" as it's redundant. We can have an option of gethistory which displays incoming and outcoming instead of amount, like:

`revault-cli gethistory`:\

kind | fee | amount
-- | -- | --
cancel | 1000 | -0
spend | 1000 | -1999000
deposit | 0 | +2000000


`revault-cli gethistory -v` (or something better, idk)

kind | fee | incoming amount | outgoing_amount
-- | -- | -- | --
cancel | 1000 | 99000 | 0
spend | 1000 | 10000 | 1989000
deposit | 0 | 0  | 0

I'm not sure if we should show in the GUI how much for the change etc, is it useful for the user? 
Also, does the cancel have an "incoming amount"? It should be 0?

edouardparis: Yes, you are right it is confusing for the cancel if there is no `amount moved`
Agree for:
| kind    | fee  | amount   |
| ------  | ---- | -------  |
| cancel  | 1000 |      0    |
| spend   | 1000 | 1999000 |
| deposit |    0  | 2000000 |

danielabrozzoni: I would say with "+" and "-", no?


kind | fee | amount
-- | -- | --
cancel | 1000 | 0
spend | 1000 | -1999000
deposit | 0 | +2000000

danielabrozzoni (approved):
`test_concurrent_spends` is a bit flaky :( I've re-run the CI.

ACK f3216826d0354db8187b21c769e4627941e32fe3

doc/API.md
```diff
@@ -368,6 +369,38 @@ disregarded for forward compatibility.
...
+| -------------- | ------ | ------------------------------------------ |
+| `events`       | array  | Array of [Event resource](#event-resource) |
+
+##### Event Resource
+
+| Field    | Type   | Description                                                                              |
+| -------- | ------ | ---------------------------------------------------------------------------------------- |
+| `kind`   | string | Type of the event, can be `deposit`, `cancel`, `spend`                                   |
+| `date`   | int    | Timestamp of the event                                                                   |
+| `amount` | int    | Outgoing or incoming Amount of the event transaction, `0` if the event is a `cancel`     |
```
darosior, line 402: May i suggest:
> Absolute amount, in sats, that is entering or exiting the wallet.

Which implies:
- that a Cancel, which "stays in the wallet" has an amount null (instead of special-casing it in the doc)
- that a Spend is obviously exiting
- that a Deposit is obviously entering

doc/API.md
```diff
@@ -368,6 +369,38 @@ disregarded for forward compatibility.
...
 
+### `gethistory`
+
+`gethistory` retrieves a paginated list of accounting events.
+
+The new deposit created by a change output is not included in the list
+of events.
+
+The `amount` of a `spend` event resource does not include the value of
+the change output.
```
darosior, line 380: I wonder how we'll be able to rule out deposits that are a change..

Also, i'm not fond of special casing specific change outputs. How about:
> Aiming at giving an accounting point of view, the amounts returned by this call are the total of inflows and outflows net of any change amount (that is technically a transaction output, but not a cash outflow).

edouardparis: Then we should maybe add the txid in the event resource, so the client can find change output amount itself

darosior: I don't understand

darosior:
ping @edouardparis can we settle on this one so we can move on with the implementation?

danielabrozzoni:
ACK 45b4b29a12f7ce7abaf3c0a2772911e728dfe421

PR #208, opened 2021-04-28
Update the Cosigning Server to be able to poll previous sigs

darosior:
This is based on #202 and updates `revault_net`, `coordinatord` and `cosignerd` to:
1. Move to the serialization of messages containing an `id`
2. Be able to poll a Cosigning Server twice (ie no more crash on a communication error when polling Cosigning Servers that results in a retry)
3. Strip the signatures from the PSBT when polling the Cosigning Servers. This slightly increases the maximum number of participants in a setup where large Spend are expected. However, the Spend announcement is now largely the bottleneck. See https://github.com/revault/revaultd/issues/207

Useful to review this PR is the implementation of this behaviour in the Cosigning Server https://github.com/revault/cosignerd/pull/20

PR #210, opened 2021-04-28
Take descriptors in the configuration

### LINKED ISSUE #108: Configuration and Miniscript descriptors
Right now, we take the descriptors parameters from the config (xpubs, unvault csv) in order to create a Miniscript descriptor on first startup.
This is kind of a bridge between the pre-Miniscript era where users were used to just put xpubs and It Works :tm: and the Miniscript era as if it's not first start we'll refuse to start with a config that does not correspond to the Miniscript descriptor stored in DB.

Is it the right approach ? At the end of the reflexion i believe this is a question of how modular we would like to be. It's just that we could store the Miniscript descriptor in the config directly as they are expected to be generated in advance (during the ceremony).
It would also be less regular-user friendly, redundant with the database, and we can't just change the descriptors and get everything to work immediately..

So the question(s) to answer here is:
- Do we want to keep a configuration with a list of xpubs and parameters (such as the Unvault CSV value), or should we move to store the Miniscript descriptors right in the config ?

darosior:
Abusing the compiler to generate the Miniscript descriptors was only a temporary solution until we figure out how we are going to generate the descriptors. Turns out we have not figured out much and are just going to generate them in advance and let them be duplicated between the config and the DB (i don't think it's that bad actually). The GUI wrapper could then take care of generating them for the user with a setup wizard.

This PR:
- Is based on #208 
- Removes our reliance on the xpubs from the config
- Updates `revault_tx` to benefit from the "smart" xpub getters from the descriptors
- Removes the scattered xpubs from the config altogether in favour of taking three descriptors (Deposit, Unvault, Cpfp) 

Fixes #108


darosior:
@edouardparis can you skim through it rapidly to tell me if it's reasonable from the GUI point of view?

doc/DEMO.md
```diff
@@ -245,6 +252,47 @@ cargo build
...
...CFH8m3bnUFXWXxKVQjMXqMiQWYRhcTeZCW1QghmkNeGkPFwADfFNt9JMuW38MnYVSAV9eyqJ3A61kbsfC5PSCdkZWi7pD2L4jv6edaPxKp","xpub6FEZyiJxqwu7zkqqVGXVbGhcAj1L5imn55VVa4Mk5WE46xdAKgD4uSR9ems9EehAApZPVXFrxcLQ7zPqYywu1z4Cjhesyp7HeRSgSdUq1BB","xpub6CFH8m3bnUFXa6UB8KhTMTPDz3cNQ9wAsQ9fGM52WZ1jBenPtf1GED6fJoDmpEYQQkk3VUHFN5ZRDLV7SRgX4M8KMTpRTH9zGRzg5udqwwo","xpub6ETaaosT68a6mxPp1dRh1yeGjMZonsZMxA1SA95iqKWDcQUXxQPYFyottUR58E8qjjnAwPcEtYS9iejkERbnGuNqfF2wgToLcxzf97FHevs"]' '["02644cf9e2b78feb0a751e50502f530a4cbd0bbda3020779605391e71654dd66c2","03ced55d1208bd8c6b42b11e29baa577711cae831b3a1296607c5e5d3ed365f49c","026237f655f3bf45fd6b7aa00e91c2603d6155f1cc001e40f5e47662d965c4c779","030a3cbcfbfdf7122fe7fa830354c956ea6595f2dbde23286f03bc1ec0c1685ca3"]' '["xpub6CFH8m3bnUFXvS78XZyCQ9mCbp7XmKXbS67YHGUS3NxHSLhAMCGHGaEPojcoYt5PYnocyuScAM5xuDzf4BqFQt3fhmKEaRgmVzDcAR46Byh","xpub6ECZqYNQzHkveSWmsGh6XSL8wMGXRtoZ5hkbWXwRSVEyEsKADe34dbdnMob1ZjUpd4TD7no1isnnvpQq9DchFes5DnHJ7JupSntZsKr7VbQ"]' 2 '["xpub6BhQvtXJmw6hi2ALFeWMi9m7G8rGterJnMTNRqUm29uvB6dVTELvnEs7hfxyN3JM48FR2oh4t8chsvw7bRRRukkyhqp9WZD4oB9UvxAMpqC","xpub6BhQvtXJmw6hksh9rRRfdLjaWjQiNMZWtkM5ebn8QkAgh5na2Un6mCDABwkUmHhPCMYtsM9zHY5jxbQ86ayvjfY8XtavbovB6NcNy8KyQLa"]' 18
```
danielabrozzoni, line 280: We've been using `cargo run` for the whole tutorial, I would use it as well too (or at least modify the other points where it was used)

danielabrozzoni: Also, this is quite complicated to read as it's all displayed on a single row. Can you either
1. Somehow split it into multiple rows or
2. Put some placeholders instead of the actual xpubs, something like:
`'["first_stakeholder_xpub", "second_stakeholder_xpub"]'` etc

doc/DEMO.md
```diff
@@ -245,6 +252,47 @@ cargo build
...
+cargo build
+```
+
+This small command line tool takes 6 parameters as input, allowing to tweak the configuration of the
+Revault setup at the chain level. These tweakable parameters are:
+- The xpubs of the stakeholders
+- The raw public keys of the cosigning servers (as many as stakeholders xpubs)
+- The xpubs of the managers, and the threshold of the multisig (eg you may have 3 xpubs and a
+  threshold of 2 in order to allow 2 managers to Spend while the third one is on vacation)
+- The xpubs of the managers' "CPFP wallet". This functionality isn't implemented yet so you can put
```
danielabrozzoni, line 271: "Put garbage here" is not very helpful for the user :joy: Let's make the user generate one CPFP key to be put here?

doc/DEMO.md
```diff
@@ -264,37 +312,10 @@ echo $PWD/stake_1_data
...
-
-```
-cosigners_keys = [
-    *first cosigner key*,
-    *second cosigner key*,
-]
-```
-Please make sure that the keys are in the right order, the first as first and the second as second. Not doing so could lead to tricky bugs.
-
+- Update the `scripts_config`: replace the `deposit_descriptor`, `unvault_descriptor` and
```
danielabrozzoni, line 315: Update it for manager 1 and stakeholder 2 too, please

contrib/config_regtest.toml
```diff
@@ -7,71 +7,35 @@ data_dir = "/path/to/your/datadir/revault"
...
-# This option MUST NOT be changed after running revaultd for the first time. If you want to change it, please remove the following files:
-# - /path/to/your/data_dir/revaultd.sqlite3
-# - /path/to/your/data_dir/revaultd-watchonly-wallet-1
-unvault_csv = 18
+# The specifications of the Bitcoin Script that we are going to be tracking onchain, put here your own that you can
+# generate with the `mscompiler` tool (in `contrib/tools`).
+# These MUST NOT be changed after running revaultd for the first time, or you'll have to re-generate the database.
+# If you have to change it, be sure to remove the previous db at `/path/to/your/data_dir/network/revaultd.sqlite3`.
+[scripts_config]
+deposit_descriptor = "wsh(multi(1,xpub6BhQvtXJmw6hi2ALFeWMi9m7G8rGterJnMTNRqUm29uvB6dVTELvnEs7hfxyN3JM48FR2oh4t8chsvw7bRRRukkyhqp9WZD4oB9UvxAMpqC/*,xpub6BhQvtXJmw6hksh9rRRfdLjaWjQiNMZWtkM5ebn8QkAgh5na2Un6mCDABwkUmHhPCMYtsM9zHY5jxbQ86ayvjfY8XtavbovB6NcNy8KyQLa/*))#4s76hpqg"
```
danielabrozzoni, line 15: I would still include the xprivs here

doc/DEMO.md
darosior: Well, no ? They use the same which they copied already?

contrib/config_regtest.toml
darosior: Ok i had to regenerate everything since i didn't save the xprivs :sob: . Done.

doc/DEMO.md
darosior: I think it's valuable as a "real" usage example, what to put is already detailed before and even given by the tool:
```
Usage: '/home/darosior/projects/revault/d/contrib/tools/mscompiler/target/debug/mscompiler <stakeholders xpubs> <cosigs pubkeys> <managers xpubs> <managers threshold> <cpfp xpubs> <Unvault CSV>'
 All values are as JSON.
```
So either we keep it as such or we get rid of it imo.

Updated to use `cargo run`..

danielabrozzoni:
It looks good :) Obviously the GUI can't start ATM (it reads revaultd configuration and expects to find `stakeholder_xpubs`), maybe we should wait to have the PR for updating the GUI and then merge both together?

darosior:
Thanks, corrected.

danielabrozzoni:
Flaky CI :( Re-running

darosior:
Hope it won't anymore post https://github.com/revault/revaultd/pull/215

danielabrozzoni (approved):
ACK 3b6316262a00f8e70b663a4565c8bc022817ada7

PR #211, opened 2021-05-03
Fix unvault txs retrieval from database 

danielabrozzoni:


darosior (approved):
ACK 709ca09922339320648932b326a20475fb1e7ee8

The test succesfully reproduces the crash without the fix:
> [1620140688][revaultd][ERROR] panic occurred at line 307 of file src/daemon/database/interface.rs

PR #213, opened 2021-05-05
Add filter to listspendtxs

### LINKED ISSUE #121: filter `listspendtxs` result
[`listspendtxs` is upcoming]
We could filter by vault outpoints, or spend txids.. To be discussed here first to implementing it :)

danielabrozzoni:
This PR adds a filter to the `listspendtx` call, meant to be used to improve the general UI/UX in the home screen. I propose to create three different sections in the home regarding spends:
1. Pending spends, which contains the spends which are awaiting signatures, and the ones not broadcasted yet (`final`, `non_final`)
2. Ongoing spends, which contains the spends we already knew, in the states `Unvaulting` `Unvaulted` `Spending` (they map to the status `Ongoing` and `Broadcasted` statuses in this PR. I left the `Ongoing` and `Broadcasted` statuses separated as they could be useful, but for the purpose of this PR I might as well merge them)
3. Unknown spends (or a better name), which contains the spends we didn't know, aka every vault in the state of `Unvaulting`, `Unvaulted`, `Spending`, but for which we didn't know the spend tx. This adds a little bit of complexity (what if we knew the spend, but a different one was broadcasted? How do we notice?) and I'm not really sure it can be made, but it would be cool to have.

Fixes #121

### DIFF
--- a/contrib/tools/psbtsigner/src/main.rs
+++ b/contrib/tools/psbtsigner/src/main.rs
@@ -10,31 +10,46 @@ use revault_tx::{
     miniscript::descriptor::DescriptorSecretKey,
 };
 
-fn parse_args(args: Vec<String>) -> (PrivateKey, Psbt, usize) {
+fn parse_args(args: Vec<String>) -> (Vec<PrivateKey>, Psbt, usize) {
     if args.len() < 3 {
         eprintln!(
-            "Usage: '{} <xpriv/derivation/path OR wif_privkey> <psbt> [input index]'",
+            "Usage: '{} <xpriv/derivation/path OR wif_privkey> [<xpriv/derivation/path OR wif_privkey>, ...] <psbt> [input index]'",
             args[0]
         );
         process::exit(1);
     }
 
-    let key = DescriptorSecretKey::from_str(&args[1]).unwrap_or_else(|e| {
-        eprintln!("Error parsing private key: '{}'", e);
+    let mut keys = vec![];
+    // Parsing keys until we find something else...
+    for arg in &args[1..] {
+        let key = match DescriptorSecretKey::from_str(arg) {
+            Ok(DescriptorSecretKey::SinglePriv(single_priv)) => single_priv.key,
+            Ok(DescriptorSecretKey::XPrv(xpriv)) => {
+                let secp = secp256k1::Secp256k1::signing_only();
+                xpriv
+                    .xkey
+                    .derive_priv(&secp, &xpriv.derivation_path)
+                    .unwrap()
+                    .private_key
+            }
+            Err(_) => {
+                // This is not a key, let's go on
+                break;
+            }
+        };
+        keys.push(key);
+    }
+
+    let psbt = base64::decode(&args.get(keys.len() + 1).unwrap_or_else(|| {
+        eprintln!(
+            "Error: Missing PSBT"
+        );
+        eprintln!(
+            "Usage: '{} <xpriv/derivation/path OR wif_privkey> [<xpriv/derivation/path OR wif_privkey>, ...] <psbt> [input index]'",
+            args[0]
+        );
         process::exit(1);
-    });
-    let key = match key {
-        DescriptorSecretKey::SinglePriv(single_priv) => single_priv.key,
-        DescriptorSecretKey::XPrv(xpriv) => {
-            let secp = secp256k1::Secp256k1::signing_only();
-            xpriv
-                .xkey
-                .derive_priv(&secp, &xpriv.derivation_path)
-                .unwrap()
-                .private_key
-        }
-    };
-    let psbt = base64::decode(&args[2]).unwrap_or_else(|e| {
+    })).unwrap_or_else(|e| {
         eprintln!("PSBT is invalid base64: '{}'", e);
         process::exit(1);
     });
@@ -43,7 +58,7 @@ fn parse_args(args: Vec<String>) -> (PrivateKey, Psbt, usize) {
         process::exit(1);
     });
     let input_index = args
-        .get(3)
+        .get(keys.len() + 2)
         .map(|i| {
             i.parse::<usize>().unwrap_or_else(|e| {
                 eprintln!("Parsing input index: '{}'", e);
@@ -55,7 +70,7 @@ fn parse_args(args: Vec<String>) -> (PrivateKey, Psbt, usize) {
             0
         });
 
-    (key, psbt, input_index)
+    (keys, psbt, input_index)
 }
 
 fn sighash(psbt: &Psbt, input_index: usize) -> (SigHash, SigHashType) {
@@ -99,8 +114,10 @@ fn sign_psbt(key: PrivateKey, psbt: &mut Psbt, input_index: usize) {
 }
 
 fn main() {
-    let (privkey, mut psbt, index) = parse_args(env::args().collect());
-    sign_psbt(privkey, &mut psbt, index);
+    let (privkeys, mut psbt, index) = parse_args(env::args().collect());
+    for privkey in privkeys {
+        sign_psbt(privkey, &mut psbt, index);
+    }
 
     let raw_psbt = serialize(&psbt);
     println!("Signed PSBT:\n{}", base64::encode(raw_psbt));
--- a/doc/API.md
+++ b/doc/API.md
@@ -338,8 +338,20 @@ disregarded for forward compatibility.
 
 #### Request
 
-| Field          | Type   | Description                                         |
-| -------------- | ------ | --------------------------------------------------- |
+| Field          | Type   | Description                                                          |
+| -------------- | ------ | -------------------------------------------------------------------- |
+| `status`       | array  | Array of [Spend status](#spend_status) |
+
+##### Spend status
+
+Please note that this status refers only to the Spend transaction, with regarding to the signatures and the broadcast status.
+You'll have to manually fetch the vaults statuses if you want to know, for example, if the vault was canceled or not.
+
+| Value           | Description                                                                                                                                                                              |
+| --------------- | ------------------------------------------------------------------------------------------------ |
+| `non_final`     | The Spend transaction is not final, we are awaiting signatures either from managers or cosigners |
+| `pending`       | The transaction is not broadcasted to the Bitcoin network                                        |
+| `broadcasted`   | The Spend transaction has been broadcasted                                                       |
 
 #### Response
 
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -52,7 +52,7 @@ use std::{
     thread::JoinHandle,
 };
 
-use serde::{Serialize, Serializer};
+use serde::{Deserialize, Serialize, Serializer};
 
 /// A presigned transaction
 #[derive(Debug, Serialize)]
@@ -124,6 +124,14 @@ where
     } else {
         s.serialize_none()
     }
+}
+
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum ListSpendStatus {
+    NonFinal,
+    Pending,
+    Broadcasted,
 }
 
 /// Any error that could arise during the process of executing the user's will.
--- a/src/daemon/database/actions.rs
+++ b/src/daemon/database/actions.rs
@@ -731,6 +731,7 @@ pub fn db_mark_rebroadcastable_spend(
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::database::schema::DbSpendTransaction;
     use crate::revaultd::RevaultD;
     use common::config::Config;
     use revault_tx::{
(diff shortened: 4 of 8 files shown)


darosior:
> This adds a little bit of complexity (what if we knew the spend, but a different one was broadcasted? How do we notice?

Didn't double check, but on the top of my head it may not be that bad. Currently the logic is that we'll only try to broadcast Spends from the `spend_transaction` table if we saw them already (implies we are a manager), and we'll update the `spend_txid` when we notice an Unvault was spent (and this was neither a Cancel nor an UnvaultEmer). The latter is not influenced by the content of `spend_transaction` (as it's only available to managers that saw the tx already).
Thus, there is no issue with broadcasting a conflicting Spend transaction (there is a functional test for that, btw).

src/daemon/jsonrpc/api.rs
```diff
@@ -1021,13 +1025,61 @@ impl RpcApi for RpcImpl {
...
+                } else if vault_statuses.iter().all(|v| *v == VaultStatus::Spending) {
+                    // At least, did we broadcast it?
+                    ListSpendStatus::Broadcasted
+                } else if let Ok(()) = psbt.clone().finalize(&revaultd.secp_ctx) {
+                    ListSpendStatus::Final
+                } else {
+                    // FIXME: Maybe here the tx is final, but another manager has all the
+                    // signatures, and our local version is still non final
+                    ListSpendStatus::NonFinal
+                };
```
darosior, line 1077: Hmm.. It seems quite convoluted to me to retrieve the statuses of the vaults a Spend is unvaulting.

darosior (commented):
What is the rationale for not just having 3 statuses for the Spend as described in #121 ? Just exposing
- `ongoing` (it was set with `updatespendtx`)
- `immature` (it was signed by the Cosigning Servers and is waiting to be broadcasted, via `setspendtx`)
- `sent` or `broadcasted` (it was confirmed, after the timelock expired)

would be trivial since we already have all the information in the table:
https://github.com/revault/revaultd/blob/3f0a5b70be5c21898c0002d0fa927ae6f8eb82bf/src/daemon/database/schema.rs#L99-L111

danielabrozzoni:
Looking only at the spend_transactions table, if I have only the non_finalized PSBT (because someone else finalized it), or if I do have the finalized PSBT but I didn't broadcast it (someone else broadcasted it), do I have enough information to correctly assume the spend status?

Example:
2 managers
Manager 1 signs the PSBT and gives it to manager 2, manager 2 signs it and clicks "broadcast" on the interface
Do I have enough information in Manager's 1 db to show that the spend is "immature"?

contrib/tools/psbtsigner/src/main.rs
```diff
@@ -10,31 +10,37 @@ use revault_tx::{
...
-                .xkey
-                .derive_priv(&secp, &xpriv.derivation_path)
-                .unwrap()
-                .private_key
-        }
-    };
-    let psbt = base64::decode(&args[2]).unwrap_or_else(|e| {
+    let mut keys = vec![];
+    // Parsing keys until we find something else...
+    for i in 0.. {
```
darosior, line 24: You should upperbound it so that `args[i + 1]` doesn't overflow? Starting at `1` would also make it more readable imo. something like
```rust
for i in 1..args.len() - 1 { }
```
[...]
```rust
DescriptorSecretKey::from_str(&args[i])
```

doc/API.md
```diff
@@ -340,6 +340,15 @@ disregarded for forward compatibility.
...
 | Field          | Type   | Description                                         |
 | -------------- | ------ | --------------------------------------------------- |
+| `status`       | string | Filter the spend transactions by their [Spend status](#spend_status) |
+
+##### Spend status
+
+| Value           | Description                                                                                      |
+| --------------- | ------------------------------------------------------------------------------------------------ |
+| `non_final`     | The spend transaction is not final, we are awaiting signatures either from managers or cosigners |
+| `immature`      | We are awaiting for the `unvault_csv` to be able to broadcast the spend                          |
```
darosior, line 350: How about
> We are waiting for the Unvault timelock to mature [...]

doc/API.md
```diff
@@ -340,6 +340,15 @@ disregarded for forward compatibility.
...
+| `broadcasted`   | The spend transaction has been broadcasted                                                       |
```
darosior, line 351: nit: i try to use uppercase for Spend Unvault Deposit etc.. in new inclusions

doc/API.md
```diff
@@ -340,6 +340,15 @@ disregarded for forward compatibility.
 
 | Field          | Type   | Description                                         |
 | -------------- | ------ | --------------------------------------------------- |
+| `status`       | string | Filter the spend transactions by their [Spend status](#spend_status) |
```
darosior, line 343: Ed isn't going to be happy if the table mark is not aligned with the above :p

doc/API.md
darosior, line 351: Maybe also point here or in the introduction of the call that the status is only concerning the Spend wrt the signatures / Unvault timelock maturation and therefore they must refer to the status of each vault to know wether it was succesfully spent, canceled, or even emergencied

src/daemon/control.rs
```diff
@@ -126,6 +126,14 @@ where
     }
 }
 
+#[derive(Debug, Deserialize, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum ListSpendStatus {
+    NonFinal,
+    Immature,
+    Broadcasted,
+}
```
darosior, line 135: neat

src/daemon/database/interface.rs
```diff
@@ -633,9 +633,10 @@ impl TryFrom<&Row<'_>> for DbSpendTransaction {
 /// List all Spend transactions in DB along with the vault they are spending
 pub fn db_list_spends(
     db_path: &PathBuf,
-) -> Result<HashMap<Txid, (SpendTransaction, Vec<OutPoint>)>, DatabaseError> {
+) -> Result<HashMap<Txid, (SpendTransaction, Vec<OutPoint>, Option<bool>)>, DatabaseError> {
```
darosior, line 636: Could it return a `DbSpendTransaction` instead of a 3-members tuple?

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+    for v in vaults:
+        deposits.append(f"{v['txid']}:{v['vout']}")
+        deriv_indexes.append(v["derivation_index"])
+    man.wait_for_active_vaults(deposits)
+    spend_tx = man.rpc.getspendtx(deposits, destinations, feerate)["spend_tx"]
+
+    for man in rn.mans():
+        spend_tx = man.man_keychain.sign_spend_psbt(spend_tx, deriv_indexes)
+        man.rpc.updatespendtx(spend_tx)
+        wait_for(
```
darosior, line 307: hmm you could even directly assert this right?

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+        lambda: len(man.rpc.listspendtxs("immature")["spend_txs"]) == 1,
+    )
+
+    rn.bitcoind.generate_block(4, wait_for_mempool=len(deposits))
+
+    for w in rn.participants():
+        wait_for(
+            lambda: len(w.rpc.listvaults(["unvaulted"], deposits)["vaults"])
+            == len(deposits)
+        )
```
darosior, line 333: Fwiw i don't get the point of newlines in the middle of a "section", but i don't really care.

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+            lambda: len(w.rpc.listvaults(["unvaulting"], deposits)["vaults"])
+            == len(deposits)
+        )
+
+    # The initiator will see the spend as immature
+    wait_for(
+        lambda: len(man.rpc.listspendtxs("immature")["spend_txs"]) == 1,
+    )
+
+    rn.bitcoind.generate_block(4, wait_for_mempool=len(deposits))
```
darosior, line 327: super nit: `4` could be a variable :p

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+    for w in rn.participants():
+        wait_for(
+            lambda: len(w.rpc.listvaults(["unvaulted"], deposits)["vaults"])
+            == len(deposits)
+        )
+
+    # Still immature...
+    wait_for(
+        lambda: len(man.rpc.listspendtxs("immature")["spend_txs"]) == 1
+    )
```
darosior, line 338: i think this can be an assert

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+    for w in rn.participants():
+        wait_for(
+            lambda: len(w.rpc.listvaults(["spending"], deposits)["vaults"])
+            == len(deposits),
+        )
+
+    # Broadcasted!
+    wait_for(
+        lambda: len(man.rpc.listspendtxs("broadcasted")["spend_txs"]) == 1
+    )
```
darosior, line 351: i think this can just be an assert

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+
+    for w in rn.participants():
+        wait_for(
+            lambda: len(w.rpc.listvaults(["spent"], deposits)["vaults"])
+            == len(deposits),
+        )
+
+    wait_for(
+        lambda: len(man.rpc.listspendtxs("broadcasted")["spend_txs"]) == 1
+    )
```
darosior, line 363: same here

tests/test_misc.py
```diff
@@ -281,6 +281,100 @@ def test_listpresignedtransactions(revault_network):
...
+    for v in vaults:
+        rn.secure_vault(v)
+        rn.activate_vault(v)
+    rn.unvault_vaults_anyhow(vaults)
+    rn.cancel_vault(vaults[0])
+
+    wait_for(
+        lambda: len(man.rpc.listspendtxs("broadcasted")["spend_txs"]) == 1,
+        debug_fn=lambda: man.rpc.listspendtxs()
+    )
```
darosior, line 375: Nice to test this. Should be an assert though (and the debug_fn is probably a debugging leftover :p)

darosior (commented):
Looks good to me overall, a few comments on the specifics.

@edouardparis you might want to give the API a look

doc/API.md
darosior: Right, concept ACK as we've done that for all RPC commands

tests/test_misc.py
danielabrozzoni: Yeah should be better now

contrib/tools/psbtsigner/testing_keys.sh
```diff
@@ -0,0 +1,3 @@
+S1KEY="xprv9whdJhXXBMAUfFPpeJkSfyCjdGDcs3FiWKL6Xn4hKgpYQ6uaCnTkZhND8Cy4NwtAVhazDrcWgqLJGWc7mLt7iG2hvVKZg4sPf8pavLFJRUH"
+S2KEY="xprv9xDe7xTWoLhKtu6BcLuWqwNVUTEE6v8j5jahb75rTsQD1RctZLVHLKXNdr7snbgt7BH9vRvCNx9xdCQnSvGnwZyAZmFwTuKuDyU8S3xVvnz"
+M1KEY="xprv9wC5JsVNK9HwGcCAuZGZE5ijdV2ST232JSXvCaT7jfWrGHqZn9EJGW76XNFtxbH7TwBQR1WSzhsiG2wwudDkJZi4ic4HJfDUcwRwHKgYufK"
```
darosior, line 3: image

doc/API.md
```diff
@@ -338,8 +338,20 @@ disregarded for forward compatibility.
...
+
+##### Spend status
+
+Please note that this status refers only to the Spend transaction, with regarding to the signatures and the broadcast status.
+You'll have to manually fetch the vaults statuses if you want to know, for example, if the vault was canceled or not.
+
+| Value           | Description                                                                                                                                                                              |
+| --------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `non_final`     | The Spend transaction is not final, we are awaiting signatures either from managers or cosigners                                                                                         |
+| `pending`       | The transaction is not broadcasted to the Bitcoin network. Either we are awaiting for the Unvault timelock to mature, or the transaction was canceled and we can't broadcast it anymore. |
```
darosior, line 353: > or the transaction was canceled and we can't broadcast it anymore.

I think this is wrong. If the Spend was once broadcasted we never mark it back to `broadcastable` in the DB?

src/daemon/database/actions.rs
```diff
@@ -1249,12 +1257,19 @@ mod test {
...
+                    broadcasted: None
+                },
+                vec![outpoint]
+            ))
         );
 
         // And delete it
         db_delete_spend(&db_path, &spend_tx.txid()).unwrap();
-        assert_eq!(db_list_spends(&db_path).unwrap().get(&spend_txid), None,);
+        assert!(db_list_spends(&db_path).unwrap().get(&spend_txid).is_none());
```
darosior, line 1272: yas i speak rust vary much

darosior (commented):
Looks good to me code wise, a comment on the doc though.

contrib/tools/psbtsigner/testing_keys.sh
danielabrozzoni: Ahahah btw those were my testing keys that absolutely didn't have to be included in the commit 🤦🏻 Thanks god we're in regtest

doc/API.md
danielabrozzoni: You're right. Added a test for this

PR #215, opened 2021-05-06
refactoring: bitcoind module cleanups

darosior:
In a nutshell this:
- DRYs-up the code
- Re-organizes the module (poller thread code in poller.rs, utils.rs, farewell actions.rs)
- Fixes some intricate logic to keep "single transitions" (eg deposit -> unvaulting -> unvaulted)
- Some misc fixes and cleanups

### DIFF
--- a/src/daemon/bitcoind/interface.rs
+++ b/src/daemon/bitcoind/interface.rs
@@ -81,6 +81,63 @@ impl BitcoinD {
         "revault-unvault".to_string()
     }
 
+    // Reasonably try to be robust to possible spurious communication error.
+    fn handle_error(&self, e: jsonrpc::Error, start: Instant) -> Result<(), BitcoindError> {
+        let now = Instant::now();
+
+        match e {
+            jsonrpc::Error::Transport(ref err) => {
+                log::error!("Transport error when talking to bitcoind: '{}'", err);
+
+                // This is *always* a simple_http::Error. Rule out the error that can
+                // not occur after startup (ie if we encounter them it must be startup
+                // and we better be failing quickly).
+                let any_err = err as &dyn Any;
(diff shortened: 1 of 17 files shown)


danielabrozzoni:
Please rebase :)

darosior:
Rebased, finally :)

src/daemon/bitcoind/interface.rs
```diff
@@ -889,6 +890,19 @@ impl BitcoinD {
             Err(e) => Err(e),
         }
     }
+
+    /// Check whether a transaction is part of the wallet, and not stuck (as in is confirmed or
+    /// part of the mempool).
+    pub fn is_current(&self, txid: &Txid) -> Result<bool, BitcoindError> {
```
danielabrozzoni, line 896: I think the naming here should be improved a bit, because it's not trivial for me to understand what this function does
So it returns true if it's in mempool or in the blockchain, false otherwise? `in_wallet` maybe?

danielabrozzoni: Or is `current` a standard term that I had never heard? :)

darosior: No, i don't think it is ;) Suggestions are welcome, but i want to express "wallet tx that is either in mempool or block chain". I wanted to call it is_valid at first but it clashes with PSBTs validity so imo it's not clear. So i though about `is_currently_valid` (as in is valid according to the current tip), hence `is_current` :D

danielabrozzoni: I have no better ideas :( Let's leave it as it is, then we can always refactor

danielabrozzoni (approved):
ACK 8a9a04642985365de404fb7b05aa69b6e972c0b0

PR #218, opened 2021-06-01
Add change_index, cpfp_index to listspendtxs

danielabrozzoni:


darosior (commented):
This needs a sanity check (at least that the entries are present, and optimally that eg the change is `None` or not when it should) and a doc update.

I don't know what's best for the handling of "multiple Change/Cpfp outputs" where we'd currently give the index of the *last* one.

src/daemon/jsonrpc/api.rs
```diff
@@ -1075,9 +1076,53 @@ impl RpcApi for RpcImpl {
...
+                .enumerate()
+            {
+                if cpfp_addresses.contains(&txout.script_pubkey) {
+                    cpfp_index = Some(i);
+                } else if deposit_addresses.contains(&txout.script_pubkey) {
+                    change_index = Some(i);
+                }
+            }
+
+            let cpfp_index = cpfp_index.expect("CPFP output must be there");
```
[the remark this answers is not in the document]
danielabrozzoni: > Also, could you double check that we wouldn't accept such a transaction if given by another manager (ie for updatespendtx that the parsing of a Spend from a PSBT in revault_tx covers that)

I'm afraid we don't do that check as I can't really find it in the code and I don't have a quick way of trying it out. I'm opening an issue in revault_tx to double check, and for now I'm removing the panic if we don't have the CPFP :)

src/daemon/jsonrpc/api.rs
```diff
@@ -1075,9 +1076,53 @@ impl RpcApi for RpcImpl {
...
+                .unsigned_tx
+                .output
+                .iter()
+                .enumerate()
+            {
+                if cpfp_addresses.contains(&txout.script_pubkey) {
+                    cpfp_index = Some(i);
+                } else if deposit_addresses.contains(&txout.script_pubkey) {
+                    change_index = Some(i);
+                }
```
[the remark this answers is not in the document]
danielabrozzoni: Created #219 for further discussion

danielabrozzoni:
Now `change_indexes` and `cpfp_indexes` are vectors, this way you can have multiple changes/cpfp

src/daemon/jsonrpc/api.rs
```diff
@@ -1075,9 +1076,54 @@ impl RpcApi for RpcImpl {
...
             }
+
+            let spent_vaults = db_vaults_from_spend(&db_path, &db_spend.psbt.txid())
+                .map_err(|e| internal_error!(e))?;
+            let cpfp_addresses: Vec<_> = spent_vaults
+                .values()
+                .map(|v| {
+                    revaultd
+                        .cpfp_descriptor
+                        .derive(v.derivation_index, &revaultd.secp_ctx)
```
danielabrozzoni, line 1087: Still TODO: check multiple derivation indexes instead of just one!

danielabrozzoni:
Speaking with Antoine IRL, we decided to avoid having cpfp and change outputs as vectors - if there are multiple change/cpfp outputs we'll notice just the last one.
The way revault_tx calculates the cpfp and change address is by using the highest derivation index from the input vaults, and we do the same for noticing the cpfp/change index. This means that if the user spends `derivation_address/0` and sends to `derivation_address/5`, we won't consider this as a change output!

src/daemon/jsonrpc/api.rs
```diff
@@ -1075,9 +1076,49 @@ impl RpcApi for RpcImpl {
...
+                .global
+                .unsigned_tx
+                .output
+                .iter()
+                .enumerate()
+            {
+                if cpfp_address == txout.script_pubkey {
+                    cpfp_index = Some(i);
+                }
+
```
darosior, line 1111: So currently (and this is unlikely to change anytime soon) the CPFP output is always going to be the first one and the change output the last one:
CPFP: https://github.com/revault/revault_tx/blob/6a0c2759332e4674e7a1f288241b0a546422d4f5/src/transactions.rs#L1203-L1208
Change: https://github.com/revault/revaultd/blob/062abff12e2dc3800b2bdf21b8853adae6a6df92/src/daemon/jsonrpc/api.rs#L952-L967

It's brittle but we should try to stick to it in the edgy edge case that they have multiple CPFP outputs. Therefore:
```rust
                if cpfp_address == txout.script_pubkey {
                    if cpfp_index.is_none() {
                        cpfp_index = Some(i);
                    }
                }
```

doc/API.md
```diff
@@ -363,11 +363,14 @@ You'll have to manually fetch the vaults statuses if you want to know, for examp
...
-| `psbt`                 | string        | Base64-encoded Spend transaction PSBT                                 |
+| Field               | Type          | Description                                                          |
+| ------------------- | ------------- | -------------------------------------------------------------------- |
+| `deposit_outpoints` | string array  | Array of the deposit outpoints of the vaults this transaction spends |
+| `psbt`              | string        | Base64-encoded Spend transaction PSBT                                |
+| `change_index`      | integer       | Index of the change output, might be null                            |
+| `cpfp_index`        | integer       | Index of the CPFP outputs                                            |
 
+`change_index` and `cpfp_index` indicate the index of the change (if any) and CPFP outputs in the outputs array as created by `getspendtransaction`. This does not aim to tag all the outputs paying to either a CPFP or a Deposit descriptor, as that would be impossible to guarantee. If two outputs pay to the change or CPFP address, the index of the last one will be returned.
```
darosior, line 373: Re my comment below, don't forget to update this too

src/daemon/jsonrpc/api.rs
```diff
@@ -1075,9 +1076,49 @@ impl RpcApi for RpcImpl {
...
+            let derivation_index = spent_vaults
+                .values()
+                .map(|v| v.derivation_index)
+                .max()
+                .expect("Spent vaults should not be empty");
+            let cpfp_address = revaultd
+                .cpfp_descriptor
+                .derive(derivation_index, &revaultd.secp_ctx)
+                .into_inner()
+                .script_pubkey();
```
darosior, line 1091: nit: calling the spk an address is slightly confusing :)

src/daemon/jsonrpc/api.rs
```diff
@@ -1075,9 +1076,49 @@ impl RpcApi for RpcImpl {
...
                 }
             }
+
+            let spent_vaults = db_vaults_from_spend(&db_path, &db_spend.psbt.txid())
+                .map_err(|e| internal_error!(e))?;
+            let derivation_index = spent_vaults
+                .values()
+                .map(|v| v.derivation_index)
+                .max()
+                .expect("Spent vaults should not be empty");
```
darosior, line 1086: Did you check that assumption? It seems to me that `updatespendtx` would happily insert a Spend refering to 0 vaults in the DB!
https://github.com/revault/revaultd/blob/062abff12e2dc3800b2bdf21b8853adae6a6df92/src/daemon/jsonrpc/api.rs#L1001-L1031

So i agree we should keep this assumption but we should open an issue for the `updatespendtx` "bug" then, good catch :)

darosior: Actually since the function parameter is typed, we would never accept a Spend without input (it would fail to be parsed). And all spend inputs must be unvaults, so we good.

Leaving that here still as this could use some documentation! (not your unwrap, the updatespendtx code)

tests/test_misc.py
```diff
@@ -292,6 +292,7 @@ def test_listspendtxs(revault_network, bitcoind):
         rn.secure_vault(v)
         rn.activate_vault(v)
 
+    # _any_spend_data never creates change
```
darosior, line 295: :+1:

darosior (commented):
LGTM, just my OCD would like the CPFP index to actually be the first encountered by the loop (in practice always the first output) and the change index to always be the last one as is the current behaviour (in practice always the last output).

src/daemon/jsonrpc/api.rs
danielabrozzoni: > we would never accept a Spend without input (it would fail to be parsed)
I don't understand. `updatespendtx` takes a `SpendTransaction`, which doesn't necessarily have the inputs set: https://github.com/revault/revault_tx/blob/master/src/transactions.rs#L1178

darosior: This is what is written in the comment **just above** yours..

darosior: https://github.com/revault/revault_tx/blob/6a0c2759332e4674e7a1f288241b0a546422d4f5/src/transactions.rs#L1391-L1393

darosior (commented):
utACK 651ccb4a4636b9210d5c5882420bf9d9c3dd859a

PR #232, opened 2021-07-07
Unit testing of RPC command routines

danielabrozzoni:


src/daemon/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub fn dummy_revaultd(datadir_name: &'static str, role: UserRole) -> RevaultD {
+        let repo_root = PathBuf::from(file!())
+            .parent()
+            .unwrap()
+            .parent()
+            .unwrap()
+            .parent()
+            .unwrap()
+            .to_path_buf();
+        let datadir_path: PathBuf = [repo_root.to_str().unwrap(), "test_data", datadir_name]
```
darosior, line 29: I think we should rather use scratch datadir in the cwd as done in the `miradord` pull request.

src/daemon/utils.rs
```diff
@@ -0,0 +1,84 @@
...
+            .iter()
+            .collect();
+
+        let config = match role {
+            UserRole::Stakeholder => "stake_config.toml",
+            UserRole::Manager => "man_config.toml",
+            UserRole::ManagerStakeholder => "stake_man_config.toml",
+        };
```
darosior, line 37: How about constructing `RevaultD` structs directly? Configuration files weren't a good idea of mine :/

darosior (commented):
Hmm configuration files and datadirs in `test_data` were IMO not a good pattern and we should not perpetuate them. It's a good occasion to clean it up, info in the config files can just be parsed directly in the tests to create a `RevaultD`.

src/daemon/utils.rs
danielabrozzoni: I don't think it's a good idea to use cwd just like that - you're going to have messy, dangling directories if you stop the tests before they finish (for example if you run the wrong command). I refactored it to be similar to the one in miradord, but I use `/tmp/revaultd_test_data/` as the base directory, what do you think?

src/daemon/utils.rs
danielabrozzoni: What about pasting the file there as I am doing now instead? It's quicker for me to write, but I agree that it's not that clean

src/daemon/utils.rs
darosior: That it's not portable :/ Unfortunately it's messy but really the "less dirty" solution i could find

src/daemon/utils.rs
darosior: Ok for now, can you leave a TODO?

src/daemon/utils.rs
danielabrozzoni: I tend to forget that some people don't use linux 😅 Should be good now

src/daemon/control.rs
```diff
@@ -875,3 +875,807 @@ pub struct RpcUtils {
...
+                1,
+                &outpoint,
+                &Amount::ONE_BTC,
+                ChildNumber::from_normal_idx(0).unwrap(),
+                1,
+            )
+            .unwrap();
+        }
+
+        // First vault: Funded
```
darosior, line 1017: nit: for each we can directly assert:
```diff
diff --git a/src/daemon/control.rs b/src/daemon/control.rs
index 1867f60..944cd1c 100644
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -1025,6 +1025,13 @@ mod test {
             Some(&transactions[1].as_ref().unwrap().initial_unvault_emer),
         )
         .unwrap();
+        assert_eq!(
+            db_vault_by_deposit(&db_file, &outpoints[1])
+                .unwrap()
+                .unwrap()
+                .status,
+            VaultStatus::Funded
+        );
 
         // Second vault: Secured
         db_confirm_deposit(
@@ -1039,7 +1046,7 @@ mod test {
         .unwrap();
 
         let vaults: Vec<_> = outpoints
-            .into_iter()
+            .iter()
             .map(|o| db_vault_by_deposit(&db_file, &o).unwrap().unwrap())
             .collect();
 
@@ -1105,6 +1112,13 @@ mod test {
             &revaultd.secp_ctx,
         )
         .unwrap();
+        assert_eq!(
+            db_vault_by_deposit(&db_file, &outpoints[2])
+                .unwrap()
+                .unwrap()
+                .status,
+            VaultStatus::Secured
+        );
 
         // Third vault: Active
         db_confirm_deposit(
@@ -1199,6 +1213,13 @@ mod test {
             &revaultd.secp_ctx,
         )
         .unwrap();
+        assert_eq!(
+            db_vault_by_deposit(&db_file, &outpoints[3])
+                .unwrap()
+                .unwrap()
+                .status,
+            VaultStatus::Active
+        );
 
         vaults
             .into_iter()
```

darosior (approved):
Awesome work. I have some comments as usual but this could really get in as is. This increased unit tests coverage by a big `7%` overall.

src/daemon/control.rs
```diff
@@ -508,7 +508,7 @@ pub fn presigned_tx_sighash(
     hashtype: SigHashType,
 ) -> Result<secp256k1::Message, SigError> {
     // Presigned transactions only have one input when handled by revaultd.
-    if !tx.inner_tx().global.unsigned_tx.input.len() == 1 {
+    if tx.inner_tx().global.unsigned_tx.input.len() != 1 {
```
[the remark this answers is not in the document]
danielabrozzoni: Argh, I put this fix in this MR but actually the test that unleashed it is `control::test::test_presigned_tx_sighash` in #234. Do you want me to move it?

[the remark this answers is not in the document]
darosior: It'd be more coherent yes. But it's no big deal either now that i know how to test it :)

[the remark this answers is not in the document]
danielabrozzoni: I moved it :)

darosior:
ACK ada58e8

PR #237, opened 2021-07-20
Fix test_reorged_cancel

danielabrozzoni:


src/daemon/bitcoind/poller.rs
```diff
@@ -911,18 +911,27 @@ fn unvault_spender(
...
-    // Finally, fetch the spending transaction and assume it's a Spend.
-    if let Some(spend_txid) = bitcoind.get_spender_txid(&unvault_outpoint, &previous_tip.hash)? {
+    // Finally, fetch the spending transaction
+    if let Some(spender_txid) = bitcoind.get_spender_txid(&unvault_outpoint, &previous_tip.hash)? {
         // FIXME: be smarter, all the information are in the previous call, no need for a
         // second one.
-        if bitcoind.is_current(&spend_txid)? {
-            return Ok(Some(UnvaultSpender::Spend(spend_txid)));
+
+        // Let's double-check that we didn't fetch the cancel, nor the unemer
```
darosior, line 926: Can you be more precise as to why it makes sense? So that when we revisit we get it first glance.

> In theory (read edge cases), the Cancel and UnEmer could have not been current at the last `bitcoind` poll but could be now. Be sure to not wrongly mark a Cancel or UnEmer as a Spend!

src/daemon/bitcoind/poller.rs
```diff
@@ -911,18 +911,27 @@ fn unvault_spender(
...
+        if spender_txid == cancel_txid || Some(spender_txid) == unemer_txid {
+            // Alright, the spender is the cancel or the unemer,
+            // but we just checked and they're not current.
+            return Ok(None);
```
darosior, line 930: Hmm, the bug is that they actually were.

src/daemon/bitcoind/poller.rs
```diff
@@ -1019,6 +1028,7 @@ fn handle_spent_unvault(
             }
         }
         None => {
+            // We don't remove it from the cache, so we'll check this outpoint at the next poll
```
darosior, line 1031: Not sure but iirc there is another call site where this has to be noted?

Could you make this "normal behaviour": ie not logging as error, and saying something like:
> Could not find a current transaction spending the Unvault txo at '{}', will check again at next poll

danielabrozzoni: > Not sure but iirc there is another call site where this has to be noted?

`unvault_spender` is called only here, `git grep unvault_spender`:
```
src/daemon/bitcoind/poller.rs:fn unvault_spender(
src/daemon/bitcoind/poller.rs:    match unvault_spender(revaultd, bitcoind, previous_tip, &unvault_outpoint)? {

```

darosior (approved):
ACK git grep deprecated tests/

PR #246, opened 2021-08-05
 Lay out manager threshold support structure

### LINKED ISSUE #137: Clean support for threshold of Managers
~~Maybe as part of #108 ~~ We need to find a way to get the threshold out of the Miniscript descriptor in a general manner.

*Note to self: policy lifting as thresh.*

danielabrozzoni:
Fixes #137

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -35,13 +35,14 @@ Display general information about the current daemon state.
 
 #### Response
 
-| Field         | Type    | Description                                                     |
-| ------------- | ------- | --------------------------------------------------------------- |
-| `blockheight` | integer | Current block height                                            |
-| `network`     | string  | Answer can be `mainnet`, `testnet`, `regtest`                   |
-| `sync`        | float   | The synchronization progress as percentage (`0 < sync < 1`)     |
-| `version`     | string  | Version following the [SimVer](http://www.simver.org/) format   |
-| `vaults`      | integer | Current number of vaults (unconfirmed are included)             |
+| Field                | Type    | Description                                                     |
+| -------------------- | ------- | --------------------------------------------------------------- |
+| `blockheight`        | integer | Current block height                                            |
+| `network`            | string  | Answer can be `mainnet`, `testnet`, `regtest`                   |
+| `sync`               | float   | The synchronization progress as percentage (`0 < sync < 1`)     |
+| `version`            | string  | Version following the [SimVer](http://www.simver.org/) format   |
+| `vaults`             | integer | Current number of vaults (unconfirmed are included)             |
+| `managers_threshold` | integer | Number of managers needed for spending the `unvault_tx`         |
 
 
 ### `getdepositaddress`
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -470,7 +470,7 @@ pub enum SigError {
     InvalidLength,
     InvalidSighash,
     VerifError(secp256k1::Error),
-    MissingSignature(BitcoinPubKey),
+    NotEnoughSignatures(usize, usize),
     /// Transaction for which we check the sigs does not pass sanity checks
     InsaneTransaction,
     Tx(revault_tx::Error),
@@ -482,7 +482,13 @@ impl std::fmt::Display for SigError {
             Self::InvalidLength => write!(f, "Invalid length of signature"),
             Self::InvalidSighash => write!(f, "Invalid SIGHASH type"),
             Self::VerifError(e) => write!(f, "Signature verification error: '{}'", e),
-            Self::MissingSignature(pk) => write!(f, "Missing signature for '{}'", pk),
+            Self::NotEnoughSignatures(needed, current) => {
+                write!(
+                    f,
+                    "Not enough signatures, needed: {}, current: {}",
+                    needed, current
+                )
+            }
             Self::InsaneTransaction => write!(f, "Insane transaction"),
             Self::Tx(e) => write!(f, "Error in transaction management: '{}'", e),
         }
@@ -580,6 +586,7 @@ pub fn check_unvault_signatures(
 /// If `db_vaults` does not contain an entry for each input.
 pub fn check_spend_signatures(
     secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+    managers_threshold: usize,
     psbt: &SpendTransaction,
     managers_pubkeys: Vec<DescriptorPublicKey>,
     db_vaults: &HashMap<Txid, DbVault>,
@@ -593,6 +600,7 @@ pub fn check_spend_signatures(
     }
 
     for (i, psbtin) in psbt.psbt().inputs.iter().enumerate() {
+        let mut valid_sigs = 0;
         let sighash = psbt
             .signature_hash(i, sighash_type)
             .expect("In bounds, and we just checked it was not finalized");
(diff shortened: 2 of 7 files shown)


doc/API.md
```diff
@@ -35,13 +35,14 @@ Display general information about the current daemon state.
...
-| `version`     | string  | Version following the [SimVer](http://www.simver.org/) format   |
-| `vaults`      | integer | Current number of vaults (unconfirmed are included)             |
+| Field                | Type    | Description                                                     |
+| -------------------- | ------- | --------------------------------------------------------------- |
+| `blockheight`        | integer | Current block height                                            |
+| `network`            | string  | Answer can be `mainnet`, `testnet`, `regtest`                   |
+| `sync`               | float   | The synchronization progress as percentage (`0 < sync < 1`)     |
+| `version`            | string  | Version following the [SimVer](http://www.simver.org/) format   |
+| `vaults`             | integer | Current number of vaults (unconfirmed are included)             |
+| `managers_threshold` | integer | Number of managers needed for spending the `unvault_tx`         |
```
darosior, line 45: Oh, cool

darosior (commented):
LGTM. 2da6d8694d92b094df5a994921194768ab059646 and bca0a15f76139411ce614f377b48cb723e65fde3 should be squashed. We also need to add support for the thresh in the test framework, and have a small sanity functional test.

darosior:
Reminder: we need to make sure we don't have weird assumptions wrt storage or else, like "IF an Unvault is broadcasted AND we are a manager THEN we must have the Spend in db". This held before as it was always a N-of-N, but doesn't anymore as a manager could now have no more information than a stakeholder.

danielabrozzoni:
I've added the functional test :)

> Reminder: we need to make sure we don't have weird assumptions wrt storage or else, like "IF an Unvault is broadcasted AND we are a manager THEN we must have the Spend in db". This held before as it was always a N-of-N, but doesn't anymore as a manager could now have no more information than a stakeholder.

I looked around a bit and it *seems* to me that we never assume so, but I'll have to double-check again tomorrow... Anyways in the test I shut down revaultd while it is in the unvaulting state, and I'm checking the state with all the managers (even the one who didn't know about the spend), so we should notice if there are some issues :)

tests/test_framework/revault_network.py
```diff
@@ -50,13 +50,27 @@ def __init__(
...
+        n_stkmanagers=0,
+        csv=None,
+        managers_threshold=None,
+    ):
         """
         Deploy a revault setup with {n_stakeholders} stakeholders, {n_managers}
         managers.
         """
+        # They didn't provide it, defaults to n_managers
+        # PS: No I can't just managers_threshold=n_managers in the method's signature :(
```
darosior, line 66: héhé :)

tests/test_framework/revault_network.py
```diff
@@ -50,13 +50,27 @@ def __init__(
...
+
```
darosior, line 67: Spurious whitespace?

tests/test_misc.py
```diff
@@ -1908,6 +1909,113 @@ def test_spends_conflicting(revault_network, bitcoind):
     )
 
 
+@pytest.mark.skipif(not POSTGRES_IS_SETUP, reason="Needs Postgres for servers db")
+def test_spend_threshold(revault_network, bitcoind, executor):
+    CSV = 2016
```
darosior, line 1914: nit: we don't really need a CSV that big, it'll just slow down the test by mining blocks

tests/test_misc.py
```diff
@@ -1908,6 +1909,113 @@ def test_spends_conflicting(revault_network, bitcoind):
...
+    man = revault_network.man(0)
+
+    # Get some more funds
+    bitcoind.generate_block(1)
+
+    vaults = []
+    deposits = []
+    deriv_indexes = []
+    total_amount = 0
+    for i in range(10):
```
darosior, line 1926: For the same reason, maybe `5` vaults is enough?

tests/test_misc.py
```diff
@@ -1908,6 +1909,113 @@ def test_spends_conflicting(revault_network, bitcoind):
...
+    spend_psbt.tx.calc_sha256()
+
+    # Revaultd didn't like it
+    with pytest.raises(
+        RpcError,
+        match=f"Error checking Spend transaction signature: 'Not enough signatures, needed: {managers_threshold}, current: {managers_threshold - 1}'",
+    ):
+        man.rpc.setspendtx(spend_psbt.tx.hash)
+
+    # Killing the daemon and restart shouldn't cause any issue
```
darosior, line 1959: :+1:

tests/test_misc.py
```diff
@@ -1908,6 +1909,113 @@ def test_spends_conflicting(revault_network, bitcoind):
...
+    for i in range(10):
+        amount = random.randint(5, 5000) / 100
+        vaults.append(revault_network.fund(amount))
+        deposits.append(f"{vaults[i]['txid']}:{vaults[i]['vout']}")
+        deriv_indexes.append(vaults[i]["derivation_index"])
+        total_amount += vaults[i]["amount"]
+    revault_network.activate_fresh_vaults(vaults)
+
+    feerate = 1
+    n_outputs = random.randint(1, 3)
```
darosior, line 1935: We can also get rid of the randomness, i'm not convinced it was a good idea when i did it for test_large_spend, but for this test i see no reason really

tests/test_misc.py
```diff
@@ -1908,6 +1909,113 @@ def test_spends_conflicting(revault_network, bitcoind):
...
+            lambda: len(m.rpc.listvaults(["unvaulted"], deposits)["vaults"])
+            == len(deposits)
+        )
+
+    # We'll broadcast the Spend transaction as soon as it's valid
+    # Note that bitcoind's RPC socket may timeout if it needs to generate too many
+    # blocks at once. So, spread them a bit.
+    for _ in range(10):
+        bitcoind.generate_block(CSV // 10)
+    bitcoind.generate_block(CSV % 10 - 1)
```
darosior, line 2001: Yeah, just a lower CSV will do the trick :)

darosior (commented):
light utACK eca9619fd902a2cbc4f6c9beedc2927883185b35, will take another look tomorrow

Great test, thanks.

darosior:
Need to remove the WIP prefix too :p

tests/test_framework/revault_network.py
```diff
@@ -50,13 +50,27 @@ def __init__(
...
+        if managers_threshold == None:
+            managers_threshold = n_managers
```
darosior, line 69: Shoudl be:
```python
managers_threshold = n_managers + n_stkmanagers
```

darosior (commented):
`test_spendtx_management` failure caught a mistake :)

darosior:
Something is def going on with parallelization. Maybe we should lower the `EXECUTOR_WORKERS` value (i was a bit optimistic i think).

src/daemon/control.rs
```diff
@@ -580,6 +586,7 @@ pub fn check_unvault_signatures(
 /// If `db_vaults` does not contain an entry for each input.
 pub fn check_spend_signatures(
     secp: &secp256k1::Secp256k1<secp256k1::VerifyOnly>,
+    managers_threshold: usize,
```
darosior, line 589: The doc comment of this function should be updated. Also, to avoid duplication of signature checking code (and avoid using an externally provided input when possible) we should in the future just rebuild a SpendTransaction from this one by using `add_signature`s. Could you add a `TODO` here?

darosior (commented):
Just one additional nit

src/daemon/control.rs
darosior: Ok, the doc comment is wrong i would not normally merge this but i don't want to block it further either. Opened https://github.com/revault/revaultd/issues/253

darosior:
ACK 3ad59e8

PR #256, opened 2021-08-18
Add getserverstatus RPC call

danielabrozzoni:


doc/API.md
```diff
@@ -431,6 +432,30 @@ None; the `result` field will be set to the empty object `{}`. Any value should
...
+| -------------- | ------ | --------------------------------------------------- |
+| `coordinator`  | object | [Server status](#server-status) for the coordinator |
+| `cosigners`    | array  | Array of [Server status](#server-status)            |
+| `watchtowers`  | array  | Array of [Server status](#server-status)            |
+
+##### Server status
+
+| Field    | Type   | Description                                                 |
+| -------- | ------ | ----------------------------------------------------------- |
+| `status` | string | Status of the server, can be `reachable`, `not_reachable`   |
```
darosior, line 456: If this is a string for forward compatibility, we can just add new entries to the object without breaking parsing for previous versions. I think it should only be a `reachable` bool.

doc/API.md
```diff
@@ -431,6 +432,30 @@ None; the `result` field will be set to the empty object `{}`. Any value should
...
+| `host`   | string | Address of the server                                       |
```
darosior, line 457: nit: hostname *and port* of the server

doc/API.md
danielabrozzoni: I was thinking that in the future we could go with more detailed status, such as "it's dead", "it's alive but I can't complete the handshake", "it's alive and we can chit-chat without troubles"

danielabrozzoni:
I've actually had correct information about the watchtowers in the call... They're all dead now but soon they won't be :D

doc/API.md
darosior: Then it would be an API break. Adding fields to an object isn't

danielabrozzoni: Ah, good point! I'll update the code

danielabrozzoni:
Updated with `reachable` bool

tests/test_misc.py
```diff
@@ -2728,3 +2728,61 @@ def test_stkman_only(revault_network, bitcoind):
...
+        assert res["coordinator"]["reachable"]
+        assert res["coordinator"]["host"] == f"127.0.0.1:{rn.coordinator_port}"
+
+    # The cosigners are alive, but only the managers see them
+    for w in rn.mans():
+        res = w.rpc.call("getserverstatus")
+        for cosigner in res["cosigners"]:
+            assert cosigner["reachable"]
+            # Sadly we don't persist the cosigner ports
+            assert cosigner["host"].startswith("127.0.0.1:")
```
darosior, line 2749: How comes?? How could you even `KKTransport::connect` without the port?

tests/test_misc.py
```diff
@@ -2728,3 +2728,61 @@ def test_stkman_only(revault_network, bitcoind):
...
+    for w in rn.mans():
+        res = w.rpc.call("getserverstatus")
+        assert res["watchtowers"] == []
+
+    # Alright, let's kill everything :D
+
+    # Hacky... We actually don't want to kill the wallets,
+    # so we kill the first N + 1 servers, which happen to be the
+    # coordinator and the N cosigners
+    for server in rn.daemons[: 1 + len(rn.stks())]:
```
darosior, line 2773: Was going to say it's brittle but i don't mind hacks in tests. But it's actually wrong! https://github.com/revault/revaultd/blob/8098371d13124b51f5c340932a4b1c52c43eb5e1/tests/test_framework/revault_network.py#L220-L262

tests/test_misc.py
danielabrozzoni: Well I'm just checking if the host starts with `127.0.0.1`, then yes, there's a port after that... But I have no idea (from the test perspective) on what ports cosigners are running, so I can't check

darosior: Yeah, brainfart, sorry.

tests/test_misc.py
danielabrozzoni: It's not :) 
The first daemon is the coordinatord: https://github.com/revault/revaultd/blob/8098371d13124b51f5c340932a4b1c52c43eb5e1/tests/test_framework/revault_network.py#L197
Then, N cosigners: https://github.com/revault/revaultd/blob/8098371d13124b51f5c340932a4b1c52c43eb5e1/tests/test_framework/revault_network.py#L262, https://github.com/revault/revaultd/blob/8098371d13124b51f5c340932a4b1c52c43eb5e1/tests/test_framework/revault_network.py#L311
Then, the wallets: https://github.com/revault/revaultd/blob/8098371d13124b51f5c340932a4b1c52c43eb5e1/tests/test_framework/revault_network.py#L336

darosior: Shit, i'm definitely not woken up today.

darosior:
utACK 4b5de214fff47a42499bcc67d33c482813acc4c7

darosior:
Fuck, `test_listspendtxs` really is flaky with this ordering bullshit

Issue #141, opened 2021-03-30
Add a `help` command


danielabrozzoni:
Add a `help` command which prints all the available commands, so I don't have to open `doc/API.md` every time :D

darosior:
I think that'd be a good thing to have, could you give more specific instructions for someone to be able to start working on it? Like the expected JSON output.

danielabrozzoni:
I would say something like this:

```
{
  "commands":[
    {
      "name":"getinfo",
      "parameters":[
        
      ],
      "description":"Display general information"
    },
    {
      "name":"getrevocationtxs",
      "parameters":[
        "outpoint"
      ],
      "description":"Retrieve the Revault revocation transactions to sign"
    }
  ]
}
```

- The description should be the same as in `doc/API.md`
- Optional parameters should be enclosed in `[` `]`, for example:
```
"parameters":[
  "[outpoint]",
  "[status]",
],
```
means that both `outpoint` and `status` are optional

PR #257, opened 2021-08-22
help cmd as per #141

vaibhavgarg237:


darosior:
Can you remove 5441068 please?

src/daemon/jsonrpc/api.rs
```diff
@@ -333,6 +337,123 @@ impl RpcApi for RpcImpl {
         }))
     }
 
+    fn help(&self, meta: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value> {
+        Ok(json!(
+        {
+            "Commands":[
```
darosior, line 348: lowercase in fields please

darosior (commented):
Looks good from a quick skim, but you didn't `cargo fmt`. I wonder if we could have something more generic (eventually?).

vaibhavgarg237:
> Looks good from a quick skim, but you didn't `cargo fmt`. I wonder if we could have something more generic (eventually?).

Can you please elaborate a little more about making it more generic? ( Something like importing commands from a separate file? )

darosior:
I was more musing about the future, this is the right approach for now!

src/daemon/jsonrpc/api.rs
```diff
@@ -91,6 +91,10 @@ pub trait RpcApi {
     #[rpc(meta, name = "getinfo")]
     fn getinfo(&self, meta: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value>;
 
+    // Print all available commands
```
darosior, line 94: nit: This should be a [doc comment](https://doc.rust-lang.org/reference/comments.html) to match all other methods

src/daemon/jsonrpc/api.rs
```diff
@@ -333,6 +337,123 @@ impl RpcApi for RpcImpl {
...
 
+    fn help(&self, meta: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value> {
+        Ok(json!(
+        {
+            "commands":[
+                {
+                  "name": "getinfo",
+                  "parameters": [
+
+                  ],
```
darosior, line 348: For all empty arrays of parameters, can you put the square brackets on the same line? ie:
```json
"parameters": [],
```

src/daemon/jsonrpc/api.rs
```diff
@@ -333,6 +337,123 @@ impl RpcApi for RpcImpl {
         }))
     }
 
+    fn help(&self, meta: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value> {
```
darosior, line 340: You don't use `meta` here, yet you have to keep it in the function declaration. To indicate you don't use it (and quiet the compiler warning), please prefix the variable name with an underscore (or just use an underscore instead):
```diff
diff --git a/src/daemon/jsonrpc/api.rs b/src/daemon/jsonrpc/api.rs
index 76a4c79..e11a11d 100644
--- a/src/daemon/jsonrpc/api.rs
+++ b/src/daemon/jsonrpc/api.rs
@@ -337,7 +337,7 @@ impl RpcApi for RpcImpl {
         }))
     }
 
-    fn help(&self, meta: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value> {
+    fn help(&self, _: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value> {
         Ok(json!(
         {
             "commands":[
```

src/daemon/jsonrpc/api.rs
```diff
@@ -333,6 +337,123 @@ impl RpcApi for RpcImpl {
...
+                {
+                    "name": "getrevocationtxs",
+                    "parameters": [
+                        "outpoint"
+                    ],
+                    "description": "Retrieve the Revault revocation transactions to sign",
+                },
+                {
+                    "name":"getunvaulttx",
+                    "parameters":[
```
darosior, line 360: nit: missing a space between `:` and `[`

darosior: Same for every object below

darosior (commented):
Some more nits

vaibhavgarg237:
> Some more nits

made required changes & clubbed in single commit...

darosior:
Looks good to me, utACK 9238ecb41752e097443d0bc117df35ebd4ac932e

danielabrozzoni:
I noticed just now that it's a bit hard to navigate - that's my fault, I didn't notice when I wrote the spec for the command. For the future, maybe we should switch to having the name of the command as the key, like:
```
[
  {
    "getinfo": { ... }
  },
  {
    "getrevocationtxs": { ... }
  }
]
```

But for now it's good as it is! Thanks :)

ACK acbc03f436ca112bc8addefbc1882c80064c118e

src/daemon/jsonrpc/api.rs
```diff
@@ -333,6 +337,105 @@ impl RpcApi for RpcImpl {
...
     }
 
+    fn help(&self, _: Self::Metadata) -> jsonrpc_core::Result<serde_json::Value> {
+        Ok(json!(
+        {
+            "commands": [
+                {
+                  "name": "getinfo",
+                  "parameters": [],
+                  "description": "Display general information",
```
darosior, line 347: Just noticed when merging that indentation is off..

darosior (commented):
Also, we should probably mention it in `doc/API.md`

PR #262, opened 2021-08-31
Fix failed building wheel for psycopg2

vaibhavgarg237:
[psycopg-binary vs psycopg](https://www.psycopg.org/docs/install.html#psycopg-vs-psycopg-binary),  `psycopg2-binary` contains compiled wheel-package with binary while `psycopg2` required a bunch of [prerequisites](https://pypi.org/project/psycopg2-binary/#:~:text=Building%20Psycopg%20requires%20a%20few%20prerequisites%20(a%20C%20compiler%2C%20some%20development%20packages)%3A%20please%20check%20the%20install%20and%20the%20faq%20documents%20in%20the%20doc%20dir%20or%20online%20for%20the%20details).
Was throwing error previously while installing dependencies for _functional tests_.

Credits:- @darosior

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -4,5 +4,5 @@ pytest-xdist==1.31.0
 pytest-timeout==1.3.4
 ephemeral_port_reserve==1.1.1
 bip32==0.0.8
-psycopg2==2.8
+psycopg2-binary==2.9
 pynacl==1.4


darosior:
utACK bc052d6af1f62a79b85800f171ebeb20079cc925

Sam071100:
But this error is related to Linux env as while building wheel for psycopg2 requires `libpq-dev` package which can be easily installed by running the command `sudo apt install libpq-dev` and works fine in Windows environment.

darosior:
The `psycopg2` package requires a global installation of Postgre, and we don't want to force that on contributors when we have a `psycopg2-binary` package which neatly comes with pre-compiled Postgre utilities.

This is clearly a win, do you know of any overlooked drawback to this approach @Sam071100?

darosior:
Since @Sam071100 didn't come up with a rationale after a week and i don't see any reason for this package not to work on Windows (they even have a Windows CI!), i'm going to merge this.

PR #267, opened 2021-09-07
tests: update python-bip32 to 2.0

darosior:
After the latest fixes uncovered by the new BIP's test vectors


darosior:
Ack 65575de

PR #271, opened 2021-09-16
refac: revaultd lib

### LINKED ISSUE #270: revaultd lib: expose daemon and revaultd module + libsodium dep
for import in the gui, it needs:
- `revaultd::Revaultd`
- `main::daemon_main` -> to move to `daemon` ?
- `sodiumoxide`  maybe expose the dep in lib.rs ?

edouardparis:
change `daemon` as a module and expose it in lib.rs
expose `revault_net` and `revault_tx`
move binaries to src/bin

close #270


src/daemon/mod.rs
```diff
@@ -7,48 +7,28 @@ mod sigfetcher;
...
-        eprintln!("Unknown arguments '{:?}'.", args);
-        eprintln!("Only '--conf <configuration file path>' is supported.");
-        process::exit(1);
-    }
-
-    Some(PathBuf::from(args[2].to_owned()))
-}
-
-fn daemon_main(mut revaultd: RevaultD) {
+pub fn daemon_main(mut revaultd: RevaultD) {
```
danielabrozzoni, line 31: Is there a rationale of why you're leaving some methods here? Isn't it better to move everything to `bin`?

Cargo.toml
```diff
@@ -9,18 +9,13 @@ keywords = ["revault", "bitcoin", "vault"]
...
 
-
-[lib]
-name = "common"
-path = "src/common/lib.rs"
-
 [[bin]]
 name = "revaultd"
-path = "src/daemon/main.rs"
+path = "src/bin/daemon/main.rs"
```
danielabrozzoni, line 14: I'd call this `src/bin/daemon.rs` if possible, so it's shorter...

Cargo.toml
```diff
@@ -9,18 +9,13 @@ keywords = ["revault", "bitcoin", "vault"]
...
 
 [[bin]]
 name = "revault-cli"
-path = "src/cli/main.rs"
+path = "src/bin/cli/main.rs"
```
danielabrozzoni, line 18: ditto above

src/daemon/mod.rs
edouardparis: Yes these methods will be required by the GUI to run the daemon in a thread

danielabrozzoni: Ah right 👍🏻

darosior:
This looks good to me and will also help to satisfy my fuzzing OCD! I agree with Daniela, i hate the Java-project-like depth our source currently has let's not make this worse unless we need it (we have a single file per binary for now).

darosior:
ACK 0b5ee23023ab1b7b42e28f280f6f806cc778a5d5

danielabrozzoni (approved):
utACK 0b5ee23023ab1b7b42e28f280f6f806cc778a5d5

PR #279, opened 2021-10-15
New api method: gethistory

edouardparis:
`gethistory` retrieves the accounting events between two timestamps
The choice of removing the `limit` parameter in favor of a `lower_limit` timestamp is because of the need of retrieving a more large set of vaults in the sql query in order to be able to not count change and cancel outputs as deposit events

based on #280


darosior:
Right we should probably contribute it upstream haha

edouardparis:
[WIP]
I still need to fix some previous comments and use the block timestamp

darosior:
Check if by any chance the "received time" field in `gettransaction` is the block timestamp if the tx is confirmed

edouardparis:
Deprecated in favor of #317

PR #287, opened 2021-10-23
WIP: tests: decrease the bitcoind poll interval

darosior:
This would half the test suite duration time, but unfortunately the reorg tests are flaky with a too low poll interval. I'd like to fix that but for now i've split it from #278 which is more pressing.


darosior:
So reducing the coordinator poll interval is resulting in a lot of timeout errors when testing on signet only with 2 wallets. I think it could be the cause of a lot of errors in regtest (with an order of magnitude more wallets).

PR #302, opened 2021-10-27
Add not_so_dummy CPFP

danielabrozzoni:
# Warning: for testing this PR you need to re-generate your CPFP descriptors using mscompiler!

## Still TODO:
- [ ] Proper fee estimation! We know that core says "idk ¯\_(ツ)_/¯" quite often, let's estimate the fees by ourselves instead of doing nothing
- [ ] a `cpfp` RPC command to manually cpfp a spend/unvault, or maybe a `prioritize` to change the `priority` status is enough


src/daemon/database/interface.rs
```diff
@@ -855,3 +855,51 @@ pub fn db_vaults_from_spend(
 
     Ok(db_vaults)
 }
+
```
danielabrozzoni, line 858: FIXME: how do I merge db_transaction_has_priority and db_transaction_has_priority_dbtx? 😭

src/daemon/bitcoind/poller.rs
```diff
@@ -1536,6 +1536,16 @@ fn maybe_create_wallet(revaultd: &mut RevaultD, bitcoind: &BitcoinD) -> Result<(
         }
         log::trace!("Importing unvault descriptors '{:?}'", &addresses);
         bitcoind.startup_import_unvault_descriptors(addresses, wallet.timestamp, fresh_wallet)?;
+
+        if revaultd.our_man_xpub.is_some() {
+            // Importing the CPFP addresses as well
+            let mut addresses = revaultd.all_cpfp_addresses();
+            for i in 0..addresses.len() {
+                addresses[i] = bitcoind.addr_descriptor(&addresses[i])?;
```
darosior, line 1736: Contributing `addr()` descriptors to `rust-miniscript` (in addition to the existing `Bare`) would be a nice performance improvement

src/daemon/revaultd.rs
```diff
@@ -260,6 +299,9 @@ pub struct RevaultD {
     pub secp_ctx: secp256k1::Secp256k1<secp256k1::VerifyOnly>,
     /// The locktime to use on all created transaction. Always 0 for now.
     pub lock_time: u32,
+    /// CPFP private key to fee-bump transactions. Managers have it if they create a cpfp_secret
+    /// file containing the key. If you don't have this key, you obviously can't cpfp.
+    pub cpfp_key: Option<ExtendedPrivKey>,
```
darosior, line 304: May i suggest:

> CPFP private key to fee-bump Unvault and Spend transactions. In the absence of the key file in the data directory, the automated CPFP mechanism is silently disabled.

src/daemon/revaultd.rs
```diff
@@ -345,6 +387,35 @@ impl RevaultD {
...
+                            k.xkey == pubkey
+                        } else {
+                            // It shouldn't happen, but I'm not going to complain...
+                            false
+                        }
+                    })
+                    .ok_or(CpfpKeyError::KeyNotInDescriptor)?;
+            } else {
+                log::warn!(
+                    "CPFP key not found, consider creating a cpfp_secret file in the datadir"
```
darosior, line 411: > Automated CPFP won't be available

or something similar would help i guess

src/daemon/revaultd.rs
```diff
@@ -345,6 +387,35 @@ impl RevaultD {
         let noise_secret_file = [data_dir_str, "noise_secret"].iter().collect();
         let noise_secret = read_or_create_noise_key(noise_secret_file)?;
 
+        let cpfp_key = if our_man_xpub.is_some() {
+            let cpfp_key_file = [data_dir_str, "cpfp_secret"].iter().collect();
+            let key = read_cpfp_key(cpfp_key_file)?;
+            if let Some(key) = key {
+                // Checking if the key is in the cpfp descriptor
```
darosior, line 394: Nice :+1:

src/daemon/revaultd.rs
```diff
@@ -345,6 +387,35 @@ impl RevaultD {
...
+                let secp_ctx = secp256k1::Secp256k1::signing_only();
+                let pubkey = ExtendedPubKey::from_private(&secp_ctx, &key);
+                cpfp_descriptor
+                    .xpubs()
+                    .iter()
+                    .find(|k| {
+                        if let DescriptorPublicKey::XPub(k) = k {
+                            k.xkey == pubkey
+                        } else {
+                            // It shouldn't happen, but I'm not going to complain...
```
darosior, line 404: I don't understand this comment

src/daemon/revaultd.rs
```diff
@@ -345,6 +387,35 @@ impl RevaultD {
...
+                            false
+                        }
+                    })
+                    .ok_or(CpfpKeyError::KeyNotInDescriptor)?;
```
darosior, line 408: Once it's merged, can you open a good first issue for testing this?

src/daemon/database/schema.rs
```diff
@@ -89,6 +89,7 @@ CREATE TABLE spend_inputs (
     id INTEGER PRIMARY KEY NOT NULL,
     unvault_id INTEGER NOT NULL,
     spend_id INTEGER NOT NULL,
+    has_priority BOOLEAN CHECK (has_priority IN (0,1, NULL)),
```
darosior, line 92: +1 for the check :+1:

What if we make this just a boolean and `false` by default? A future `prioritisetransaction` RPC command could just set this to `true` and `updatespendtx` with the optional priority parameter would also updated it to `true`.

src/daemon/database/interface.rs
darosior: You can't without an invasive refactoring. You could still do that to DRY the more footguny part:
```diff
diff --git a/src/daemon/database/interface.rs b/src/daemon/database/interface.rs
index 9e03856..d4f36a0 100644
--- a/src/daemon/database/interface.rs
+++ b/src/daemon/database/interface.rs
@@ -86,6 +86,7 @@ where
     x
 }
 
+// FIXME: have a version mergeable with the "regular" db_query
 fn db_query_tx<'a, P, F, T>(
     db_tx: &Transaction,
     stmt_str: &'a str,
@@ -97,7 +98,6 @@ where
     P::Item: ToSql,
     F: FnMut(&Row<'_>) -> rusqlite::Result<T>,
 {
-    // rustc says 'borrowed value does not live long enough'
     db_tx
         .prepare(stmt_str)
         .map_err(|e| DatabaseError(format!("Preparing query: '{}'", e.to_string())))?
@@ -856,18 +856,21 @@ pub fn db_vaults_from_spend(
     Ok(db_vaults)
 }
 
-pub fn db_transaction_has_priority(db_path: &Path, txid: &Txid) -> Result<bool, DatabaseError> {
-    // We don't know if this tx is an unvault or a spend...
-    db_query(
-        db_path,
-        "SELECT has_priority \
-         FROM spend_transactions as stx \
-         INNER JOIN spend_inputs as sin ON stx.id = sin.spend_id \
-         INNER JOIN presigned_transactions as ptx ON ptx.id = sin.unvault_id \
-         WHERE stx.txid = (?1) OR ptx.txid = (?1)",
-        params![txid.to_vec()],
-        |row| Ok(row.get::<_, Option<bool>>(0)? == Some(true)),
-    )
+const TX_HAS_PRIO_QUERY: &str = "\
+SELECT has_priority
+    FROM spend_transactions as stx
+        INNER JOIN spend_inputs as sin ON stx.id = sin.spend_id
+        INNER JOIN presigned_transactions as ptx ON ptx.id = sin.unvault_id
+    WHERE stx.txid = (?1) OR ptx.txid = (?1)
+";
+
+pub fn db_transaction_has_priority_dbtx(
+    db_tx: &Transaction,
+    txid: &Txid,
+) -> Result<bool, DatabaseError> {
+    db_query_tx(db_tx, TX_HAS_PRIO_QUERY, params![txid.to_vec()], |row| {
+        Ok(row.get::<_, Option<bool>>(0)? == Some(true))
+    })
     .map(
         // if the tx is not here, no need to return an error: it doesn't have priority
         |mut v| {
@@ -879,20 +882,11 @@ pub fn db_transaction_has_priority(db_path: &Path, txid: &Txid) -> Result<bool,
     )
 }
 
-pub fn db_transaction_has_priority_dbtx(
-    db_tx: &Transaction,
-    txid: &Txid,
-) -> Result<bool, DatabaseError> {
-    db_query_tx(
-        db_tx,
-        "SELECT has_priority \
-         FROM spend_transactions as stx \
-         INNER JOIN spend_inputs as sin ON stx.id = sin.spend_id \
-         INNER JOIN presigned_transactions as ptx ON ptx.id = sin.unvault_id \
-         WHERE stx.txid = (?1) OR ptx.txid = (?1)",
-        params![txid.to_vec()],
-        |row| Ok(row.get::<_, Option<bool>>(0)? == Some(true)),
-    )
+pub fn db_transaction_has_priority(db_path: &Path, txid: &Txid) -> Result<bool, DatabaseError> {
+    // We don't know if this tx is an unvault or a spend...
+    db_query(db_path, TX_HAS_PRIO_QUERY, params![txid.to_vec()], |row| {
+        Ok(row.get::<_, Option<bool>>(0)? == Some(true))
+    })
     .map(
         // if the tx is not here, no need to return an error: it doesn't have priority
         |mut v| {
```

src/daemon/database/interface.rs
```diff
@@ -855,3 +855,51 @@ pub fn db_vaults_from_spend(
...
+pub fn db_transaction_has_priority(db_path: &Path, txid: &Txid) -> Result<bool, DatabaseError> {
```
darosior, line 859: smol doc comment for public interface plz :)

src/daemon/jsonrpc/api/mod.rs
```diff
@@ -208,6 +208,7 @@ pub trait RpcApi {
         &self,
         meta: Self::Metadata,
         spend_txid: Txid,
+        priority: bool,
```
darosior, line 211: Any reason for not making it an optional parameter? (no opinion, just asking)

Also needs a API.md update

src/daemon/bitcoind/interface.rs
```diff
@@ -33,7 +33,7 @@ const RPC_SOCKET_TIMEOUT: u64 = 180;
 // Labels used to tag utxos in the watchonly wallet
 const DEPOSIT_UTXOS_LABEL: &str = "revault-deposit";
 const UNVAULT_UTXOS_LABEL: &str = "revault-unvault";
-const CPFP_UTXOS_LABEL: &str = "revault-cpfp";
+pub const CPFP_UTXOS_LABEL: &str = "revault-cpfp";
```
darosior, line 36: meh. Didn't read the entire commit yet but it smells like a layer violation

src/daemon/bitcoind/interface.rs
```diff
@@ -897,6 +897,31 @@ impl BitcoinD {
             Ok((_, None, _)) => self.is_in_mempool(txid),
         }
     }
+
+    /// Estimates the feerate needed for a tx to make it in the
+    /// next block. Uses estimatesmartfee and, in case it returns an
+    /// error, a default value.
+    /// The value returned is in sats/WU
+    // FIXME: Can I use u64? Wtf? I'm not sure
```
darosior, line 905: No. `bitcoind`'s 1sat/vbyte is ~`0.253` sat/WU.

darosior: You can use sats/kWU to avoid that. Or just use vbytes

src/daemon/bitcoind/interface.rs
```diff
@@ -897,6 +897,31 @@ impl BitcoinD {
...
+    pub fn estimate_feerate(&self) -> Result<u64, BitcoindError> {
+        if let Ok(json) = self.make_node_request(
+            "estimatesmartfee",
+            &params!(Json::Number(serde_json::Number::from(1))),
```
darosior, line 909: Can you use `2`? internally bitcoind will translate 1 to 2.

src/daemon/bitcoind/interface.rs
```diff
@@ -897,6 +897,31 @@ impl BitcoinD {
...
+                // Math is hard
+                // btc/kvbyte -> sats/kbyte
+                let sats_kvb = btc_kvb * Amount::ONE_BTC.as_sat() as f64;
+                // sats/kbyte -> sats/vbyte
+                let sats_vb = sats_kvb / 1000.0;
+                // sats/vbyte -> sats/WU
+                return Ok((sats_vb * 4.0) as u64);
+            }
+        }
+        // TODO fallback feerate
```
darosior, line 922: Heh! We can't do that.

src/daemon/control.rs
```diff
@@ -726,10 +740,212 @@ pub fn watchtowers_status(revaultd: &RevaultD) -> Vec<ServerStatus> {
             });
         }
     }
-
     watchtowers
 }
 
+// FIXME: probably this function should be in revault_tx, not here...
```
darosior, line 746: Yes.

src/daemon/control.rs
```diff
@@ -726,10 +740,212 @@ pub fn watchtowers_status(revaultd: &RevaultD) -> Vec<ServerStatus> {
...
+fn build_cpfp_tx(
```
darosior, line 747: This definitely needs a unit test.

src/daemon/control.rs
```diff
@@ -726,10 +740,212 @@ pub fn watchtowers_status(revaultd: &RevaultD) -> Vec<ServerStatus> {
...
+            value: 0,
+            script_pubkey: Script::new_op_return(&[]),
+        }],
+    };
+    let mut psbt =
+        PartiallySignedTransaction::from_unsigned_tx(transaction).expect("We don't sign the tx");
+
+    loop {
+        // TODO: I should consider the package fees and not the individual transaction fees
+        // to be more precise
```
darosior, line 809: More than that, it's actually a requirement for this to make sense at all :sweat_smile:

darosior (commented):
First incomplete pass at reviewing the diff. Only reviewed the first 3 commits and quickly skimmed through the 4th one.

src/daemon/control.rs
danielabrozzoni: In the todo list already, working on it :)

src/daemon/control.rs
danielabrozzoni: Same as above, it's in my todo list (see the PR description) with high prio, but you can probably look at the general structure of the cpfp (the tracking etc) without this detail. That's mostly why I opened a PR with a dummy, full of TODOs CPFP.

src/daemon/bitcoind/interface.rs
danielabrozzoni: Why would we ask for two blocks? Why would bitcoind translate 2 blocks target to 1? I don't get it 😅

src/daemon/revaultd.rs
danielabrozzoni: I'm still unsure what I should do here.
When I wrote the code, I thought "I shouldn't find a non-xpub in a cpfp descriptor. If I do, I'll just ignore it". Now I'm thinking I should just panic:
```
                    .find(|k| {
                        if let DescriptorPublicKey::XPub(k) = k {
                            k.xkey == pubkey
                        } else {
                            // CPFP descriptor always contains xpubs
                            unreachable!();
                        }
                    })
```
What do you think?

src/daemon/jsonrpc/api/mod.rs
danielabrozzoni: Well `priority` is just two states: either the tx has priority or it doesn't. It seemed weird to represent this in the API with a three states scenario, where both `None` and `Some(false)` mean "no priority" and `Some(true)` means "with priority".

On the other side, the GUI needs an update if this is not optional (at first, since we don't have the UI for asking for automatic CPFP, we would just send `priority: false` all the time)

src/daemon/control.rs
danielabrozzoni: Is this a "yes, let's do it now" or a "yes, we'll move it" eventually?

src/daemon/bitcoind/poller.rs
```diff
@@ -386,23 +423,92 @@ fn mark_confirmed_emers(
...
+                        txid
+                    );
+                    return Ok(());
+                }
+                Err(RpcControlError::MissingCpfpInput) => {
+                    // We thought we never CPFPed this transaction (feebumped_blockheight is None!), but actually,
+                    // *shit happens*, and the cpfp outpoint has already been spent. Could be:
+                    // - If a transaction is dropped from the mempool we're going to re-insert
+                    // it in the cache with "feebumped_blockheight: None", even tho maybe it already has
+                    // a feebump tx
```
darosior, line 469: When a tx is dropped from the mempool, its child are as well. But yeah it's racy

src/daemon/bitcoind/poller.rs
```diff
@@ -386,23 +423,92 @@ fn mark_confirmed_emers(
...
+                    // - In a really high fee scenario, we might have used this transaction's feebump
+                    // output for feebumping another transaction, effectively cpfping this one as well
+                    log::warn!("We wanted to feebump transaction '{}', but its CPFP output is not in listunspent!", txid);
```
darosior, line 472: You could avoid these edge cases most of the case by calling `listunspent` with `minconf = 0`. You still need to handle race conditions though :/

src/daemon/bitcoind/poller.rs
```diff
@@ -386,23 +423,92 @@ fn mark_confirmed_emers(
...
+                    info.feebumped_blockheight = Some(current_blockheight);
+                    continue;
+                }
+                Err(e) => return Err(BitcoindError::Custom(e.to_string())),
+            };
+            bitcoind.broadcast_transaction(&tx)?;
```
darosior, line 478: If erroring here would have us crash, you must not error: anything could have happened that would prevent `sendrawtransaction` to return a success.

src/daemon/bitcoind/poller.rs
```diff
@@ -762,6 +887,21 @@ fn comprehensive_rescan(
...
+                        CpfpTxInfo {
+                            broadcasted_blockheight,
+                            feebumped_blockheight: None,
+                        },
+                    );
+                    log::debug!(
+                        "Inserted Unvault transaction {} in CPFPable table as it got unconfirmed (unvault case 3)",
+                        unvault_txid
+                    );
+                }
```
darosior, line 903: Oh, that's very noble but i wouldn't even bother with trying to handle CPFP with reorgs. It's something we can add later and that is going to be a large can of worms if we add now.

src/daemon/bitcoind/poller.rs
darosior: Actually i think i misunderstood your comment

danielabrozzoni: Yeah I meant: if a transaction had cpfp descendants and is dropped from the mempool and then broadcasted again we insert it again in the cache with "feebump_blockheight": None, even tho it still has descendants (we just don't know)

src/daemon/bitcoind/poller.rs
danielabrozzoni: I am calling listunspent with minconf = 0 already (otherwise I would never find the outputs I need to use to feebump, as I feebump only unconf txs 😄 ). The second scenario is not something I can really prevent... I accidentally feebumped this transaction while feebumping another one. It's fine. Actually, it's great 😄

src/daemon/bitcoind/interface.rs
danielabrozzoni: Made it a `pub(super) const`, slightly better :D I need this label in the poller to filter the cpfps in listunspent.

Or maybe i should have some method here to filter listunspent?

src/daemon/jsonrpc/api/mod.rs
danielabrozzoni: Actually, thinking more about it, it makes sense to just have an Option and default it to false. Will update!

darosior:
Cool the list of done/todo is much more reasonable now :) Just one question before i dive in the code:
>  Proper fee estimation! We know that core says "idk ¯_(ツ)_/¯" quite often, let's estimate the fees by ourselves instead of using 1 sat/WU

Is it outdated or do you do that? It's quite useless to bump by 1sat/WU so let's just do nothing without proper estimation.

danielabrozzoni:
At the moment i just use 10 sats/WU and if it's lower than the past feerate I don't do anything. The problem is that in the tests I never have proper estimation, so I hardcoded 10 sats/WU to be able to test what I have, I'm still thinking how to overcome this issue... (I'm not sure the fallback algorithm would work, but I'm going to look into it soon-ish)

src/daemon/bitcoind/poller.rs
```diff
@@ -8,6 +8,7 @@ use crate::daemon::{
         },
         BitcoindError,
     },
+    cpfp::{get_cpfp_transaction, CpfpError, CpfpTx},
```
darosior, line 11: In commit "Track and broadcast CPFP transaction when needed", you are going through a lot of hassle managing this cache. But do we need to track the CPFP transactions?

What would be wrong with an approach similar to the way we broadcast the Spend transactions (apart from being bulky):
- On new blocks, iterate through the Spend transactions that are waiting for being broadcast (as we do now)
- Store the ones that are high prio and whose Unvaults are still unconfirmed
- After the loop, create a CPFP spending all the unconfirmed Unvaults (and Spends) that were marked as priority
- Take care to check if there is already a CPFP tx in your mempool, if so increase the fees accordingly to replace it (more absolute fees, an added feerate >1sat/vbyte)

And ta-da: you don't need all the complexity of managing the cache (which currently fails in some cases as you outline in the commit message) and you even get RBF for free. Am i missing something?

tests/test_chain.py
```diff
@@ -511,7 +511,7 @@ def test_retrieve_vault_status(revault_network, bitcoind):
     spend_psbt = serializations.PSBT()
     spend_psbt.deserialize(spend_tx)
     spend_psbt.tx.calc_sha256()
-    mans[0].rpc.setspendtx(spend_psbt.tx.hash)
+    mans[0].rpc.setspendtx(spend_psbt.tx.hash, False)
```
darosior, line 514: In commit " tests: Pass priority in setspendtx and... ...in broadcast/spend utilities"

Why is it needed if you made the field optional on the RPC interface?

darosior (commented):
Ok, that was a big undertaking :)

I have quite a few comments, the most important one being wrt the CPFP tx tracking. Overall i like the organization, and this PR was easy to follow :+1: .

I did not yet review the functional tests but here are the comments already. I'll look into mocking the bitcoind interface in the functional tests.

src/daemon/revaultd.rs
```diff
@@ -260,6 +299,10 @@ pub struct RevaultD {
     pub secp_ctx: secp256k1::Secp256k1<secp256k1::VerifyOnly>,
     /// The locktime to use on all created transaction. Always 0 for now.
     pub lock_time: u32,
+    /// CPFP private key to fee-bump Unvault and Spend transactions.
+    /// In the absence of the key file in the data directory, the automated CPFP mechanism
+    /// is silently disabled.
```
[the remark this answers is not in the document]
danielabrozzoni: https://github.com/revault/revaultd/pull/302#discussion_r742990987 lol

src/daemon/bitcoind/interface.rs
```diff
@@ -572,44 +575,74 @@ impl BitcoinD {
...
+            Some(UNVAULT_UTXOS_LABEL),
+        )
+    }
+
+    pub fn list_unspent_cpfp(
+        &self,
+        min_amount: Option<u64>,
+    ) -> Result<Vec<ListUnspentEntry>, BitcoindError> {
+        // For some weird reason, listunspent with the cpfp wallet doesn't return the label
+        // (maybe because we only have one descriptor anyways?), so we pass `None` as a label
```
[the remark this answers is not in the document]
danielabrozzoni: No, it's returning utxos without the "label" attribute at all

src/daemon/bitcoind/interface.rs
```diff
@@ -1032,3 +955,88 @@ pub struct SyncInfo {
...
+                        .map(|d| ChildNumber::Normal { index: d })
+                        .ok();
+                }
+            }
+        }
+
+        ListUnspentEntry {
+            outpoint: OutPoint {
+                txid,
+                vout: vout as u32, // Bitcoin makes this safe
```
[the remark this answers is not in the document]
danielabrozzoni: I guess because of the maximum number of inputs by consensus? (Or maybe it's just a policy? Lol)
But I admit I copy-pasted the code from some other point in revaultd

src/daemon/bitcoind/poller.rs
```diff
@@ -388,23 +428,138 @@ fn mark_confirmed_emers(
...
+            let current_feerate = bitcoind.estimate_feerate()?;
+            let tx_feerate = info.tx.max_feerate() * 1000; // to sats/kWU
+            if current_feerate < tx_feerate {
+                // Uhm, we don't need to cpfp this for now, our estimate is lower
+                // than the tx fees.
+                info.feebumped_blockheight = Some(current_blockheight);
+            }
+            let added_feerate = current_feerate - tx_feerate;
+
+            let mut listunspent: Vec<_> = bitcoind.list_unspent_cpfp(None)?;
```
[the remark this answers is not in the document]
danielabrozzoni: Why does it assume confirmed utxos?

src/daemon/bitcoind/poller.rs
```diff
@@ -388,23 +428,138 @@ fn mark_confirmed_emers(
...
+                    CpfpTxIn::new(l.outpoint, txout)
+                })
+                .collect();
+
+            let tx_derived_cpfp_descriptor = revaultd
+                .derived_cpfp_descriptor(my_listunspent.derivation_index.expect("Must be here"));
+
+            log::info!("CPFPing transaction with id '{}'", txid);
+
+            use revault_tx::error::{Error, TransactionCreationError};
```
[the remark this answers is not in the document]
danielabrozzoni: I put them here on purpose, to avoid importing it globally, but matching on the error 10 rows below

src/daemon/bitcoind/poller.rs
```diff
@@ -1719,15 +2005,28 @@ pub fn poller_main(
...
     Ok(())
 }
+
+#[derive(Debug)]
+struct CpfpTxInfo {
+    // When was this tx broadcasted?
+    broadcasted_blockheight: u32,
+    // Last time we broadcasted the cpfp for this transaction, or checked it for RBF.
+    feebumped_blockheight: Option<u32>,
+    // The tx to be feebumped. Might not be there, in which case we retrieve it
```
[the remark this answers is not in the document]
danielabrozzoni: Argh old comment, sorry

tests/test_chain.py
danielabrozzoni: Forgot to update after API update 😭

src/daemon/bitcoind/poller.rs
```diff
@@ -1538,6 +1540,64 @@ fn maybe_create_wallet(revaultd: &mut RevaultD, bitcoind: &BitcoinD) -> Result<(
...
+                    log::error!("Unloading wallet '{}': '{}'", &cpfp_wallet_path, e);
+                }
+            }
+
+            bitcoind.createwallet_startup(cpfp_wallet_path, false)?;
+            log::info!("Importing descriptors to bitcoind cpfp wallet.");
+
+            // Now, import descriptors.
+            // Importing the CPFP descriptor should be cool and painless :) But it's not :)
+            // FIXME: maybe this code should be moved somewhere else
```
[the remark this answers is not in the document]
darosior: But then, should we silently temper with the descriptors the user provided? It makes more sense to me that we should just error if the descriptors aren't with testnet xpubs on testnet, but for demos and testing it's just going to be a fking nuisance

src/daemon/revaultd.rs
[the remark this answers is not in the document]
darosior: FAIR ENOUGH

src/daemon/bitcoind/interface.rs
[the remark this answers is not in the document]
darosior: Yeah i noticed :)
Oh, i commented not on the right one. Yeah this one obviously because of the size limit.
My question was intended to be about another part where we assume block heights wont be >i32::MAX because so does bitcoind.

src/daemon/bitcoind/interface.rs
```diff
@@ -445,6 +445,24 @@ impl BitcoinD {
         Ok(desc_wo_checksum)
     }
 
+    pub fn get_new_cpfp_address(&self) -> Result<Address, BitcoindError> {
+        let address = &self
+            .make_cpfp_request(
+                "getnewaddress",
+                &params!(Json::String(CPFP_UTXOS_LABEL.to_string())),
```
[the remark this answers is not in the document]
darosior: This works internally with an "address book" which tbh i'm not familiar with the intrinsic of :)

src/daemon/bitcoind/interface.rs
```diff
@@ -922,6 +991,34 @@ impl BitcoinD {
...
+                let btc_kvb = n.as_f64().expect("feerate is f64");
+                // Math is hard
+                // btc/kvbyte -> sats/kbyte
+                let sats_kvb = btc_kvb * Amount::ONE_BTC.as_sat() as f64;
+                // sats/kbyte -> sats/vbyte
+                let sats_vb = sats_kvb / 1000.0;
+                // sats/vbyte -> sats/WU
+                let sats_wu = sats_vb * 4.0;
+                // sats/WU -> msats/WU
+                return Ok((sats_wu * 1000.0) as u64);
```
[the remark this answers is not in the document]
darosior: It's because i was wrong! btc/kVB -> sat/kWU isn't just a simple /4!

Here and in general i agree with you on detailing steps.

src/daemon/bitcoind/interface.rs
```diff
@@ -922,6 +991,34 @@ impl BitcoinD {
...
+            }
+        }
+        // TODO: Calculate the fallback feerate using the blockchain!
+        // We fallback to 10 sats/WU, which is absolutely
+        // arbitrary, but makes tests happy :)
+        Ok(10_000)
```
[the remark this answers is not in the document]
darosior: I don't think it makes sense. And a bump of 1sat/WU is insignificant but would still cost significant absolute fees for nothing.

If there is no estimate, let's just do nothing for now.

src/daemon/cpfp.rs
```diff
@@ -0,0 +1,108 @@
...
+        .cpfp_transaction(
+            derived_cpfp_descriptor,
+            added_feerate,
+            listunspent,
+            change_address,
+        )?
+        .as_psbt_string();
+
+    let (complete, psbt_signed) = bitcoind.sign_psbt(psbt)?;
+    assert!(complete);
```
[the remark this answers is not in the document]
darosior: > If the CPFP descriptor is a proper cpfp descriptor (which I guess is checked), and if the cpfp key we have is contained in the descriptor (checked at startup time) this holds

and if there is no single small miscalculation wrt fees on both our *and bitcoind* side, and if bitcoind isn't going to fail to complete for whatever reason. In general, let's assert for critical stuff or when we get into an unrecoverable state: i consider CPFP an optional feature and we could just loudly log as `error` and not crash if one of the above condition is not met.
(ofc the caller should log, not this function)

src/daemon/bitcoind/poller.rs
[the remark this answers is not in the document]
darosior: Because you call `listunspent` with a minconf of `0` iirc?

src/daemon/bitcoind/poller.rs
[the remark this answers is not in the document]
darosior: Oh right, i didn't even care to look for an explanation. Then i don't really care. Is it a common Rust practice? Looked awkward to me.

src/daemon/bitcoind/interface.rs
```diff
@@ -922,6 +991,34 @@ impl BitcoinD {
...
+        ) {
+            if let Some(n) = json.get("feerate") {
+                let btc_kvb = n.as_f64().expect("feerate is f64");
+                // Math is hard
+                // btc/kvbyte -> sats/kbyte
+                let sats_kvb = btc_kvb * Amount::ONE_BTC.as_sat() as f64;
+                // sats/kbyte -> sats/vbyte
+                let sats_vb = sats_kvb / 1000.0;
+                // sats/vbyte -> sats/WU
+                let sats_wu = sats_vb * 4.0;
```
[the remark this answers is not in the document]
danielabrozzoni: > since a vbyte is 4 times larger than a weight unit

So, if you have 1sat/vbyte, how many sat/WU do you have? I'd say 4?

src/daemon/bitcoind/poller.rs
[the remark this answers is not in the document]
danielabrozzoni: That wasn't my question :) Why does `revault_tx`'s `get_cpfp_transaction` assume confirmed utxos?

darosior:
Because if you have an unconfirmed parent you need to take it into account in the package size calculation.

src/daemon/bitcoind/poller.rs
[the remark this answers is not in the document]
danielabrozzoni: I think it is, for avoiding too long branches in the match. Not sure tho

darosior:
>  a cpfp RPC command to manually cpfp a spend/unvault, or maybe a prioritize to change the priority status is enough

Maybe this could just be a parameter to `updatespendtx`?

danielabrozzoni:
Mh, I don't like it that much, `updatespendtx` is meant to be used for updating the actual spend with more signatures before broadcast, and prioritize should be set at broadcast time or afterwards, so I think a new api would be cleaner.
Anyways, I'll open an issue for the remaining todos when the PR is merged, so we can discuss it there :)

darosior (commented):
I have a number of remarks. I've made them a followup so we don't get stuck on this another two months. See #332 .

Some nits here aren't part of my followup.

ACK 2699a1b26975c875da04706229e2614f986ee1e0 -- not based on the code but because of the followup in #322.

src/daemon/database/interface.rs
```diff
@@ -859,3 +861,47 @@ pub fn db_vaults_from_spend(
...
+            };
+            let spend_id = row.get::<_, i64>(6)?;
+            unvaults
+                .entry(spend_id)
+                .or_insert(Vec::new())
+                .push(unvault_tx);
+            Ok(())
+        },
+    )?;
+    Ok(unvaults.values().cloned().collect())
```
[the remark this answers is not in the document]
danielabrozzoni: Not available on Rust 1.43

src/daemon/bitcoind/poller.rs
```diff
@@ -391,6 +393,159 @@ fn mark_confirmed_emers(
...
+        // We check if any unvault in this package is still unconfirmed. If so, we feebump
+        if unvault_package.iter().any(|u| {
+            bitcoind
+                .get_wallet_transaction(&u.txid())
+                .map(|w| w.blockheight.is_none())
+                .unwrap_or(true)
+        }) {
+            cpfp_package(revaultd, bitcoind, unvault_package, current_feerate)?;
+        }
+    }
```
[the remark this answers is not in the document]
danielabrozzoni: I don't have a valid rationale, I had reasons in the past for doing so, when refactoring the code for the tenth time in a row I overlooked that those assumptions don't hold anymore. Sorry.

Can you please stop posting multiple comments for the same problem? It's the second comment stating this same thing (and I could probably find many more examples in this same PR). If you rethink your statement after clicking the `Add comment` just click edit...

tests/test_misc.py
```diff
@@ -226,3 +227,92 @@ def test_no_cosig_server(revault_network):
...
+            lambda: len(w.rpc.listvaults(["unvaulting"])["vaults"]) == len(vaults),
+        )
+
+    # Uh oh! The feerate is too low, miners aren't including our transaction...
+    bitcoind.generate_blocks_censor(1, unvault_txids)
+    man.wait_for_log(
+        f"CPFPed transactions",
+    )
+    for unvault_txid in unvault_txids:
+        assert bitcoind.rpc.getmempoolentry(unvault_txid)["descendantcount"] == 2
```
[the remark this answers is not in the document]
darosior: Done in #332

tests/test_misc.py
```diff
@@ -226,3 +227,92 @@ def test_no_cosig_server(revault_network):
...
+
+    # Alright, now let's do everything again for the spend :tada:
+
+    # Confirming the unvaults
+    bitcoind.generate_block(1)
```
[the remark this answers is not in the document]
darosior: Done in #332

tests/test_misc.py
```diff
@@ -226,3 +227,92 @@ def test_no_cosig_server(revault_network):
...
+    bitcoind.generate_block(CSV - 1)
+    man.wait_for_log(
+        f"Succesfully broadcasted Spend tx '{spend_txid}'",
+    )
+
+    # Uh oh! The feerate is too low, miners aren't including our transaction...
+    bitcoind.generate_blocks_censor(1, [spend_txid])
+    man.wait_for_log("Checking if transactions need CPFP...")
+    # Nah, they don't
+    assert bitcoind.rpc.getmempoolentry(spend_txid)["descendantcount"] == 1
```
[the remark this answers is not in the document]
darosior: Done in #332

src/daemon/bitcoind/poller.rs
```diff
@@ -391,6 +393,159 @@ fn mark_confirmed_emers(
...
+    revaultd: &Arc<RwLock<RevaultD>>,
+    bitcoind: &BitcoinD,
+) -> Result<(), BitcoindError> {
+    let db_path = revaultd.read().unwrap().db_file();
+    log::debug!("Checking if transactions need CPFP...");
+
+    if revaultd.read().unwrap().cpfp_key.is_none() {
+        log::warn!("We should CPFP transactions, but we don't have a cpfp key!");
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
darosior: Done in #332

src/daemon/revaultd.rs
danielabrozzoni: https://github.com/revault/revaultd/issues/335

PR #308, opened 2021-11-01
Watchtower revocation signatures exchange

darosior:
This pull request, based on #306, implements the distribution of revocation signatures to the watchtower (https://github.com/revault/miradord).

The protocol defines a `sig` message per transaction and we leverage this on the watchtower to not allocate fee-bumping reserves before a vault is delegated. In order to do this we share the second-stage (ie the txs spending the Unvault) transactions signatures before sharing the Unvault to implicitly signal that we are about to share the Unvault transaction signature and the watchtower better be telling us not to do it now if it doesn't have sufficient reserve [0].
Of course, this is all simulation as we don't have fee-bumping reserve management implemented on the watchtower yet.
This translates here by having the background signature fetcher thread share the Emergency transaction signatures when it gathered them all (or to do it immediately in `revocationtxs` if we were passed them all at once).

This also adds a few plugins to test some basic spending policies (revault everything, revault nothing, revault according to a maximum per day). Note that by default we set the "revault nothing" policies as by default the watchtower would currently revault everything and this would break most of our tests :).

[0] I think we should not rely on this trick but instead modify the protocol. See https://github.com/revault/practical-revault/issues/109

### DIFF
--- a/src/daemon/jsonrpc/api/error.rs
+++ b/src/daemon/jsonrpc/api/error.rs
@@ -14,12 +14,19 @@ impl From<CommunicationError> for Error {
     fn from(e: CommunicationError) -> Error {
         let code = match e {
             CommunicationError::Net(_) => ErrorCode::TRANSPORT_ERROR,
+            CommunicationError::WatchtowerNack(_, _) => ErrorCode::WT_SIG_NACK,
             CommunicationError::SignatureStorage => ErrorCode::COORDINATOR_SIG_STORE_ERROR,
             CommunicationError::SpendTxStorage => ErrorCode::COORDINATOR_SPEND_STORE_ERROR,
             CommunicationError::CosigAlreadySigned => ErrorCode::COSIGNER_ALREADY_SIGN_ERROR,
             CommunicationError::CosigInsanePsbt => ErrorCode::COSIGNER_INSANE_ERROR,
         };
         Error(code, e.to_string())
+    }
+}
+
+impl From<CommunicationError> for JsonRpcError {
+    fn from(e: CommunicationError) -> JsonRpcError {
+        Error::from(e).into()
     }
 }
 
@@ -96,6 +103,8 @@ pub enum ErrorCode {
     INTERNAL_ERROR = 11000,
     /// An error internal to revault_net, generally a transport error
     TRANSPORT_ERROR = 12000,
+    /// The watchtower refused our signatures
+    WT_SIG_NACK = 13_000,
     /// The Coordinator told us they could not store our signature
     COORDINATOR_SIG_STORE_ERROR = 13100,
     /// The Coordinator told us they could not store our Spend transaction
--- a/src/daemon/jsonrpc/api/mod.rs
+++ b/src/daemon/jsonrpc/api/mod.rs
@@ -11,8 +11,8 @@ use crate::daemon::{
         announce_spend_transaction, check_spend_transaction_size, coord_share_rev_signatures,
         coordinator_status, cosigners_status, fetch_cosigs_signatures, finalized_emer_txs,
         listvaults_from_db, onchain_txs, presigned_txs, share_unvault_signatures,
-        vaults_from_deposits, watchtowers_status, wts_share_emer_signatures, ListSpendEntry,
-        ListSpendStatus, RpcUtils,
+        vaults_from_deposits, watchtowers_status, wts_share_emer_signatures,
+        wts_share_second_stage_signatures, ListSpendEntry, ListSpendStatus, RpcUtils,
     },
     database::{
         actions::{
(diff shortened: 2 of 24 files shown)


darosior:
Arg, small typo in the CI config to build the watchtower. Should be fixed now.

darosior:
Rebased on master

darosior:
Ok so the policy test is a bit flaky, can't manage to reproduce locally... It's ready for review nonetheless!

tests/test_rpc.py
```diff
@@ -1108,25 +1108,21 @@ def test_getserverstatus(revault_network, bitcoind):
...
         assert res["cosigners"] == []
 
-    # Only stakeholders have info about the watchtowers...
+    # Only stakeholders have info about the watchtowers, they must be reachable
     for w in rn.stks():
         res = w.rpc.call("getserverstatus")
-        # ...And well, they're dead
         for watchtower in res["watchtowers"]:
-            assert not watchtower["reachable"]
+            assert watchtower["reachable"]
```
danielabrozzoni, line 1115: 😄

tests/test_rpc.py
```diff
@@ -1108,25 +1108,21 @@ def test_getserverstatus(revault_network, bitcoind):
...
-    # Alright, let's kill everything :D
-
-    # Hacky... We actually don't want to kill the wallets,
-    # so we kill the first N + 1 servers, which happen to be the
-    # coordinator and the N cosigners
-    for server in rn.daemons[: 1 + len(rn.stks())]:
-        server.stop()
+    # Alright, let's kill all the servers
+    for d in rn.daemons:
+        if d not in rn.participants():
```
danielabrozzoni, line 1124: Ah, cool

darosior:
ACK, fwiw i have been extensively testing this on signet for the past day(s).

src/daemon/jsonrpc/api/mod.rs
```diff
@@ -745,10 +746,26 @@ impl RpcApi for RpcImpl {
         let rev_txs = vec![cancel_db_tx, emer_db_tx, unvault_emer_db_tx];
         db_update_presigned_txs(&db_path, &db_vault, rev_txs.clone(), secp_ctx)?;
         db_mark_securing_vault(&db_path, db_vault.id)?;
+
+        // If this made the Emergency fully signed and we are a stakeholder, share
+        // it with our watchtowers.
+        let emer_db_tx =
+            db_emer_transaction(&db_path, db_vault.id)?.ok_or(JsonRpcError::internal_error())?;
+        if !db_vault.emer_shared && emer_db_tx.psbt.unwrap_emer().is_finalizable(secp_ctx) {
+            if let Some(ref watchtowers) = revaultd.watchtowers {
```
danielabrozzoni, line 755: Oh wait, so are we ok with stakeholders not having watchtowers?

darosior:
> I think it might be that pytest tries to run miradord tests as well. Did it ever occur to you?

Yeah, i specify the files to run. For instance `pytest tests/test_*.py`. But a [quick search](https://stackoverflow.com/a/41940247) would give you the more appropriate `--ignore` functionality.

src/daemon/jsonrpc/api/error.rs
```diff
@@ -96,6 +103,8 @@ pub enum ErrorCode {
     INTERNAL_ERROR = 11000,
     /// An error internal to revault_net, generally a transport error
     TRANSPORT_ERROR = 12000,
+    /// The watchtower refused our signatures
+    WT_SIG_NACK = 13_000,
```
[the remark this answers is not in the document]
darosior: Because i find it much more readable and i didn't want to increase the diff with unrelated changes by modifying the others (as i did for instance to add `From` to be able to use `?` directly for some errors in #306 but didn't change all occurrences).

src/daemon/jsonrpc/api/mod.rs
darosior: Yes, i see no reason to not accept it and at least for now we need it

tests/test_framework/revault_network.py
```diff
@@ -537,10 +537,11 @@ def activate_fresh_vaults(self, vaults):
         for j in act_jobs:
             j.result(TIMEOUT)
 
-    def unvault_vaults(self, vaults, destinations, feerate):
+    def broadcast_unvaults(self, vaults, destinations, feerate):
```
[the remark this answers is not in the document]
darosior: Yes, but let's introduce it when we actually need it.

tests/test_framework/wtplugins/max_value_per_day.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+"""A plugin which enforces a maximum total value per day.
+
+It needs as part of its config:
+    - A "data_dir" entry specifying where it is going to store its 'database'.
+    - A "max_value" entry specifying the maximum value per day to enforce.
+
+It simply stores a counter which is reset to 0 after 144 blocks (assumes no reorg).
```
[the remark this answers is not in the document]
darosior: Yes it a random testing plugin (we won't put plugins in `test_framework/`, let alone in this repo..). Using system time includes a lot of non-determinism for the tests

darosior:
Added one more `wait_for_log` to try to prevent the `wt_policy` test timeout on CI which i was unable to reproduce locally.

Added a commit to run `test_watchtower` with `test_misc` to reduce CI running time since we can only have up to 4 instances running concurrently.

danielabrozzoni:
> Yeah, i specify the files to run. For instance pytest tests/test_*.py. But a quick search would give you the more appropriate --ignore functionality.

Ah, got it! I thought I was doing something wrong 😅 
Maybe update `tests/README.md` with the miradord download instructions and this `--ignore` functionality then :)

danielabrozzoni:
ACK 1c4dee062fa0d2c5ca5808ae9e2fdac2dcce97c1

darosior:
>  Maybe update tests/README.md with the miradord download instructions and this --ignore functionality then :)

Good call, done.

danielabrozzoni:
You also need:
```
diff --git a/tests/README.md b/tests/README.md
index f01f9c2..1b5c0f2 100644
--- a/tests/README.md
+++ b/tests/README.md
@@ -61,6 +61,7 @@ git submodule update --init --recursive
 cd tests/servers
 cd coordinatord && cargo build
 cd cosignerd && cargo build
+cd miradord && cargo build
 # TODO: add the other servers here when they are implemented
```

darosior:
:sob: done. Thanks for the sharp look!

danielabrozzoni:
ACK 81c532dd0f80b956f07676018f829a3ef5fff223

PR #313, opened 2021-11-02
Accept signet addresses for emergency

danielabrozzoni:
Signet addresses are parsed as testnet


src/common/config.rs
```diff
@@ -243,7 +243,9 @@ impl Config {
 
             let emer_addr_net = stk_config.emergency_address.address().network;
             let bitcoind_net = config.bitcoind_config.network;
-            if emer_addr_net != bitcoind_net {
+            // All good, signet addresses have testnet type
+            if emer_addr_net == Network::Testnet && bitcoind_net == Network::Signet {
+            } else if emer_addr_net != bitcoind_net {
```
darosior, line 248: This empty `if` clause really looks awkward to me. Also it's wrong as we still want to check for testnet.
I suggest:
```rust
// Testnet addresses are used for signet
if emer_addr_net != bitcoind_net && !(bitcoind_net == Network::Signet && emer_addr_net == Network::Testnet)
```

danielabrozzoni: > Also it's wrong as we still want to check for testnet.

We are still checking for testnet, as you wouldn't enter the first if the bitcoind_net wasn't signet.
I agree the empty if is awkward, but it seems easier to understand to me. Anyways, I can change it, no problem

darosior: Oh, right. Yeah my version is more verbose and less immediately clear..

darosior: Maybe:
```rust
let signet_special_case = bitcoind_net == Network::Signet && emer_addr_net == Network::Testnet;
if emer_addr_net != bitcoind_net && !signet_special_case
```

danielabrozzoni: Cool!

darosior:
ACK b3171df522b396be5a6a6a680f853ad2fdaa73f1

PR #321, opened 2021-12-01
panic hook: panic payload might be a String

danielabrozzoni:
The Rust documentation says that "[the PanicInfo payload] will
commonly, but not always, be a &'static str or String"

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -144,17 +144,18 @@ pub fn setup_panic_hook() {
             .unwrap_or_else(|| "'unknown'".to_string());
 
         let bt = backtrace::Backtrace::new();
-        if let Some(s) = panic_info.payload().downcast_ref::<&str>() {
-            log::error!(
-                "panic occurred at line {} of file {}: {:?}\n{:?}",
-                line,
-                file,
-                s,
-                bt
-            );
-        } else {
-            log::error!("panic occurred at line {} of file {}", line, file);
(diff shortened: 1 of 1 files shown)


src/daemon/mod.rs
```diff
@@ -146,7 +146,15 @@ pub fn setup_panic_hook() {
...
-                "panic occurred at line {} of file {}: {:?}\n{:?}",
+                "panic occurred at line {} of file {}: {}\n{:?}",
+                line,
+                file,
+                s,
+                bt
+            );
+        } else if let Some(s) = panic_info.payload().downcast_ref::<String>() {
+            log::error!(
+                "panic occurred at line {} of file {}: {}\n{:?}",
```
darosior, line 157: nit: couldn't you use smth like `unwrap_or_else()` rather than duplicating the log message?

darosior (commented):
Concept ACK, good catch. I assumed that downcasting to `&str` would work for both `&static str` and `String`. From a quick search looks like it's wrong. Maybe you want to handle `&static str` too then?

Also, you have a CI error

danielabrozzoni:
> Maybe you want to handle &'static str too then?

I think that a `&'static str` is just a `&str` with a certain lifetime, isn't it? It should be handled already.

darosior (commented):
utACK 55381386efeeb4f6ad3cd01272a9e3a8905dd6ee

PR #323, opened 2021-12-03
Unflake the watchtower policy test

darosior:

PR #325, opened 2021-12-04
qa: mockable RPC commands for `bitcoind`

darosior:
It's really helpful to have a mockable interface between the daemons and the `bitcoind` they query.
This is particularly helpful to test the error handling code for robustness by simulating some spurious failures on the `bitcoind` backend.
It also comes handy in cases like for #302 where we'd like to customize the fee estimation returned by `bitcoind` in order to test different CPFP scenarii.

This adds an optional proxy between the `revaultd`, `miradord` and their `bitcoind`. We use it to either pass through or mock commands.

Based on #323, since, well, it fixed the CI.

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -150,6 +150,7 @@ def revaultd_stakeholder(bitcoind, directory):
     coordinator_noise_key = (
         "d91563973102454a7830137e92d0548bc83b4ea2799f1df04622ca1307381402"
     )
+    bitcoind_cookie = os.path.join(bitcoind.bitcoin_dir, "regtest", ".cookie")
     revaultd = StakeholderRevaultd(
         datadir,
         dep_desc,
@@ -158,7 +159,8 @@ def revaultd_stakeholder(bitcoind, directory):
         os.urandom(32),
         coordinator_noise_key,
         reserve(),
-        bitcoind,
+        bitcoind.rpcport,
+        bitcoind_cookie,
         stk_config=stk_config,
         wt_process=None,
     )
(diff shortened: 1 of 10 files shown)


darosior:
Reviewers: i tested the pass-through with the following patch:
```patch
diff --git a/tests/test_framework/revault_network.py b/tests/test_framework/revault_network.py
index b52d51e..e88531c 100644
--- a/tests/test_framework/revault_network.py
+++ b/tests/test_framework/revault_network.py
@@ -79,7 +79,8 @@ class RevaultNetwork:
         # Connection info to bitcoind. Change the port depending on whether we are proxying
         # the daemons' requests.
         bitcoind_cookie = os.path.join(self.bitcoind.bitcoin_dir, "regtest", ".cookie")
-        if len(bitcoind_rpc_mocks) > 0:
+        # if len(bitcoind_rpc_mocks) > 0:
+        if True:
             self.bitcoind_proxy = BitcoindRpcProxy(
                 self.bitcoind.rpcport, bitcoind_cookie, bitcoind_rpc_mocks
             )
```

But i didn't test mocks.

danielabrozzoni (approved):
ACK e315737ad237fd56cdb5f97ab9795fe4ce78a0a3 - I skimmed through the code (didn't review extensively) and tested

Implemented the mock on top of CPFP code like:
```
index 69ac4b3..d4909bf 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -232,7 +232,13 @@ def test_no_cosig_server(revault_network):
 def test_cpfp_transaction(revault_network, bitcoind):
     CSV = 12
-    revault_network.deploy(2, 1, csv=CSV, with_watchtowers=False)
+    revault_network.deploy(
+        2,
+        1,
+        csv=CSV,
+        with_watchtowers=False,
+        bitcoind_rpc_mocks={"estimatesmartfee": {"feerate": 0.0005}}, # 50 sats/vbyte
+    )
     man = revault_network.mans()[0]
     stks = revault_network.stks()
     amount = 0.24
```
and it works 🎉

PR #330, opened 2021-12-12
qa: always persist logs

darosior:
Even when we raise (happens in wait_for_logs for instance)


danielabrozzoni:
utACK 3fbe66d4046f7c9c01ef3eb5fb4f1105a6400761

PR #331, opened 2021-12-13
Rework help command

danielabrozzoni:

PR #332, opened 2021-12-13
CPFP fixups and followups

darosior:
This is a followup to #302, fixing inconsistencies and cleaning up a bit.
The revault_tx counterpart is at https://github.com/revault/revault_tx/pull/117 .


tests/test_misc.py
```diff
@@ -229,41 +229,59 @@ def test_no_cosig_server(revault_network):
...
+
 @pytest.mark.skipif(not POSTGRES_IS_SETUP, reason="Needs Postgres for servers db")
 def test_cpfp_transaction(revault_network, bitcoind):
     CSV = 12
     revault_network.deploy(
         2,
         1,
         csv=CSV,
-        with_watchtowers=False,
+        # with_watchtowers=False,
```
danielabrozzoni, line 253: Don't comment, either you set to true or to false

danielabrozzoni: I don't recall any real reason, probably I followed a "if watchtowers are not needed here why spawning them at all" kinda reasoning to make the test run quicker

src/daemon/bitcoind/poller.rs
```diff
@@ -392,30 +399,75 @@ fn mark_confirmed_emers(
...
+
+// CPFP a bunch of transactions, bumping their feerate by at least `target_feerate`.
+// `target_feerate` is expressed in sat/kWU.
+// All the transactions' feerate MUST be below `target_feerate`.
 fn cpfp_package(
     revaultd: &Arc<RwLock<RevaultD>>,
     bitcoind: &BitcoinD,
-    mut tx_package: Vec<impl CpfpableTransaction>,
-    current_feerate: u64,
+    mut tx_package: Vec<ToBeCpfped>,
```
danielabrozzoni, line 444: Can we call this a package?

tests/test_misc.py
darosior: Makes sense. My own rationale for making it default to `True` was: "test in as many scenarii as possible".

src/daemon/bitcoind/poller.rs
darosior: Not really, i fix it elsewhere IIRC. If not here, the rationale was that i didn't want to increase the diff by too much.

darosior:
Fixed your comment on the tests.

darosior:
Rebased on master for Github aesthetics

danielabrozzoni (commented):
ACK 5ae472c2c995cd3297e1249e07e287fe3e9ddf3e

I'll re-ack if you decide to set the revault_tx reference to master :)

danielabrozzoni:
Uhm I just saw https://github.com/revault/revaultd/pull/302#discussion_r767336107, I'll fix in #331 after this is merged

darosior:
Yeah i need to update revault_net first

Sent from ProtonMail mobile

darosior:
Update revault_tx and revault_net to their respective master

darosior:
Spurious HTTP timeout when generating blocks on CI: added a commit (unrelated to this PR, but hey CI won't be flaky then) making the timeout much larger so we are now fine with generating loads of blocks at once.

PR #337, opened 2021-12-15
Add affected vault outpoints in gethistory response

edouardparis:
Because the client may want to have the full detail of a history event, a new field (list of affected vault outpoints) is added to gethistory event response, then the client can use `listonchaintransactions`, and have full detail of all the transactions.
It helps also the client to distinguish deposit events with the same txid and amount.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -439,14 +439,15 @@ of inflows and outflows net of any change amount (that is technically a transact
 
 ##### Event Resource
 
-| Field         | Type   | Description                                                        |
-| ------------- | ------ | ------------------------------------------------------------------ |
-| `blockheight` | int    | Blockheight of the event final transaction                         |
-| `txid`        | string | Hex string  of the event final transaction id                      |
-| `kind`        | string | Type of the event. Can be `deposit`, `cancel`, `spend`             |
-| `date`        | int    | Timestamp of the event                                             |
-| `amount`      | int    | Absolute amount in satoshis that is entering or exiting the wallet |
-| `fee`         | int    | Fees caused by the operation, includes CPFP outputs amount         |
+| Field         | Type         | Description                                                                      |
+| ------------- | ------       | -------------------------------------------------------------------------------- |
+| `blockheight` | int          | Blockheight of the event final transaction                                       |
+| `txid`        | string       | Hex string  of the event final transaction id                                    |
+| `kind`        | string       | Type of the event. Can be `deposit`, `cancel`, `spend`                           |
+| `date`        | int          | Timestamp of the event                                                           |
+| `amount`      | int          | Absolute amount in satoshis that is entering or exiting the wallet               |
+| `fee`         | int          | Fees caused by the operation, includes CPFP outputs amount                       |
+| `vaults`      | string array | List of outpoints of vaults affected by the event excluding any change vault     |
 
 
 ### `emergency`
--- a/src/daemon/control.rs
+++ b/src/daemon/control.rs
@@ -904,6 +904,7 @@ pub fn get_history<T: BitcoindThread>(
                 amount: Some(vault.amount.as_sat()),
                 fee: None,
                 txid: vault.deposit_outpoint.txid,
+                vaults: vec![vault.deposit_outpoint],
             });
         }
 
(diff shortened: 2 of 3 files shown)


doc/API.md
```diff
@@ -439,14 +439,15 @@ of inflows and outflows net of any change amount (that is technically a transact
...
-| `fee`         | int    | Fees caused by the operation, includes CPFP outputs amount         |
+| Field              | Type         | Description                                                                      |
+| -------------      | ------       | -------------------------------------------------------------------------------- |
+| `blockheight`      | int          | Blockheight of the event final transaction                                       |
+| `txid`             | string       | Hex string  of the event final transaction id                                    |
+| `kind`             | string       | Type of the event. Can be `deposit`, `cancel`, `spend`                           |
+| `date`             | int          | Timestamp of the event                                                           |
+| `amount`           | int          | Absolute amount in satoshis that is entering or exiting the wallet               |
+| `fee`              | int          | Fees caused by the operation, includes CPFP outputs amount                       |
+| `concerned_vaults` | string array | List of vault outpoints moved or created by the event excluding any change vault |
```
darosior, line 450: I think you mean "involved" (translation of "concerné"). But in this can can just be `vaults` IMO.

> List of outpoints of vaults affected by this event

src/daemon/control.rs
```diff
@@ -1060,6 +1075,8 @@ pub struct HistoryEvent {
     pub amount: Option<u64>,
     pub fee: Option<u64>,
     pub txid: Txid,
+    #[serde(serialize_with = "serialize_tostr_vec")]
```
darosior, line 1078: Why is this necessary? We've never used it for other lists of outpoints to return.

darosior (commented):
I guess this doesn't hurt. Can you update the functional test sanity checking the API as well?
I have a couple comments regarding 1. the name of the field and 2. the serialization.

tests/test_rpc.py
```diff
@@ -1198,6 +1198,7 @@ def test_gethistory(revault_network, bitcoind, executor):
     revault_network.activate_vault(canceled_vault)
     revault_network.spend_vaults_anyhow_unconfirmed([canceled_vault])
     revault_network.cancel_vault(canceled_vault)
+    canceled_outpoint = f"{canceled_vault['txid']}:{canceled_vault['vout']}"
 
     # We wait a little before and after the timestamp to keep everything
     # cleanly seperated
```
darosior, line 1204: Oh, not introduced here but it's outdated.

tests/test_rpc.py
```diff
@@ -1241,33 +1256,50 @@ def test_gethistory(revault_network, bitcoind, executor):
     ]
     assert len(events) == 2
     assert events[0]["kind"] == "cancel"
+    assert events[0]["vaults"][0] == canceled_outpoint
```
darosior, line 1259: nit: worse asserting on the `len()` too here and below.

darosior (commented):
utACK c0530df82173336e2195c28eec74877f60eba4ee

darosior (commented):
Thanks for taking the nits.
utACK af30444

doc/API.md
```diff
@@ -439,14 +439,15 @@ of inflows and outflows net of any change amount (that is technically a transact
...
-| `fee`         | int    | Fees caused by the operation, includes CPFP outputs amount         |
+| Field         | Type         | Description                                                                      |
+| ------------- | ------       | -------------------------------------------------------------------------------- |
+| `blockheight` | int          | Blockheight of the event final transaction                                       |
+| `txid`        | string       | Hex string  of the event final transaction id                                    |
+| `kind`        | string       | Type of the event. Can be `deposit`, `cancel`, `spend`                           |
+| `date`        | int          | Timestamp of the event                                                           |
+| `amount`      | int          | Absolute amount in satoshis that is entering or exiting the wallet               |
+| `fee`         | int          | Fees caused by the operation, includes CPFP outputs amount                       |
+| `vaults`      | string array | List of outpoints of vaults affected by the event excluding any change vault     |
```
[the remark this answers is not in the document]
darosior: @edouardparis you resolved the comment without answering the question... I guess my guess was right?  *<- this is a question too please don't ignore it :)*

PR #342, opened 2021-12-17
CPFP: just store the BIP32 seed, not the whole xpriv

darosior:


danielabrozzoni:
Ping @JSwambo that might need to be aware of this for the ceremony: to be able to CPFP, a manager now has to give the BIP32 seed for the CPFP xpriv, not the xpriv itself. I don;t know it this changes something on how you generate/store secrets :) 

ACK f9ca8c5b1983b765bc56b37934e8d63272d99f91 - I tested locally, a test is failing, but I believe it's not this PR's fault as it's happening on master as well (see issue #343)

PR #344, opened 2021-12-18
Library modules overhaul

darosior:
This is a cleanup of the source tree organization.

First, this splits the `control.rs` module into control and communication.

Then, this cleans up the library organization.
We previously made `revaultd` a library in order for `revault-gui` to be usable as a single binary. However, we didn't go the complete way toward a library and we ended up with a half-binary half-library source tree with more depth than needed, unneeded modules and needlessly exposed functions (why did we expose a `setup_panic_hook` routine??). This fixes it by making the choice to go through the library path. Note this had no effect on the size of the two binaries.

A followup should expose the commands as a library in addition to the current JSONRPC-via-UDS way.


src/lib.rs
```diff
@@ -32,9 +32,39 @@ use std::{
     io::{self, Write},
     panic, process,
     sync::{mpsc, Arc, RwLock},
-    thread, time,
+    thread,
 };
 
+// A panic in any thread should stop the main thread, and print the panic.
+fn setup_panic_hook() {
```
edouardparis, line 39: Maybe it is a mistake, but I thought you wanted to move it to main.rs like `setup_logger`

darosior: Why? It's used by and related to the `daemon_main` function. Do you think it should be moved to the binary? If so, why?

edouardparis: Binaries embedding revaultd, may want to use daemon_main with their own panic hook. But I am ok to keep it here, It removes code on the GUI side :+1:

darosior: I think that this is something internal to `revaultd`: we make the assumption that we can panic inside the threads we ourselves start. If someone sets the hook to something else, it might cause some weird behaviour (for instance an onchain state that doesn't get updated anymore as the bitcoind thread panic'ed and it didn't stop the main thread).

In fact, it's IMO one the most brittle things about making revaultd a library: a user of the library can rewrite this hook and by doing so would unknowingly break our assumption.

darosior:
Since this is a monster rebase i plean to do this post #333, FWIW.

darosior:
Finally rebased this. Phew.

danielabrozzoni:
re-ACK 3e320d4c17668daa009c4a8c7b2ce0a2b8f77280

I'll rebase #352 on top of this one :D

PR #345, opened 2021-12-20
Fix test_batched_cpfp_transaction

danielabrozzoni:
The test wrongly assumed that the spends would have been confirmed
in the same block, and thus cpfped in the same block.

### DIFF
--- a/src/daemon/bitcoind/poller.rs
+++ b/src/daemon/bitcoind/poller.rs
@@ -611,6 +611,9 @@ fn maybe_cpfp_txs(
         .collect();
 
     // TODO: std transaction max size check and split
+    // FIXME: to_cpfp might include some transactions that aren't confirmed, but
+    // have already been CPFPed. If that's the case, cpfp_package will fail and
+    // won't CPFP any transaction.
     if !to_cpfp.is_empty() {
         cpfp_package(revaultd, bitcoind, to_cpfp, current_feerate)?;
     } else {
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -377,7 +377,7 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
 
     # At this point, one of the first unvaults gets mined, but not the other. Feerate
     # spikes and makes us feebump: we'll create a CPFP tx spending the remaining unconfirmed
-    # Unvault from the first batch, and the 3 unvaults of the second spend.
+    # Unvault from the first batch, and the 2 unvaults of the second spend.
     revault_network.bitcoind_proxy.mocks["estimatesmartfee"] = {
         "feerate": 30 * 1_000 / COIN  # 30 is 6sats/vb above, should trigger CPFP
     }
@@ -396,14 +396,24 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
     for txid in unvaults:
         assert txid in cpfp_entry["depends"]
 
-    # Now get to be able to broadcast the second Spend.
-    bitcoind.generate_block(CSV, wait_for_mempool=unvaults)
+    # Now get to be able to broadcast the second Spend. Note that both Spend transactions
+    # are only broadcastable at the last block, but revaultd might be lagging behind: it
+    # is possible it polls the tip at block CSV-1 and actually manage to broadcast a Spend
+    # since bitcoind is already at block CSV. It would then trigger CPFP for the Spend(s)
+    # alone when mining the last block, and fail the test later on when we assert that we
+    # CPFP the two spends and the unvaults at once.
+    # Therefore we make sure revaultd is sync'ed before we mine the last block.
+    height = man.rpc.getinfo()["blockheight"]
+    bitcoind.generate_block(CSV - 1, wait_for_mempool=unvaults + [cpfp_txid])
+    wait_for(lambda: man.rpc.getinfo()["blockheight"] == height + CSV - 1)
+    bitcoind.generate_block(1)
     man.wait_for_logs(
         [
             f"broadcasted Spend tx '{first_spend}'",
             f"broadcasted Spend tx '{second_spend}'",
         ]
     )
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 2)
 
     # In the meantime, we attempt a third Spend.
     third_spend_psbt = revault_network.broadcast_unvaults_anyhow(


src/daemon/bitcoind/poller.rs
```diff
@@ -611,6 +611,9 @@ fn maybe_cpfp_txs(
         .collect();
 
     // TODO: std transaction max size check and split
+    // FIXME: to_cpfp might include some transactions that aren't confirmed, but
+    // have already been CPFPed. If that's the case, cpfp_package will fail and
+    // won't CPFP any transaction.
```
darosior, line 616: This is in essence the same FIXME as the above:
>   // FIXME: it's a shame to have to get the derivation paths from bitcoind, but we need to
>   // for the Spend CPFP output derivation index. We have all the info and should be able to
>  // make it work without this hack.

darosior: But yea it didn't mention RBF (because i wanted to have #333 merged before but hey)

darosior (commented):
> This wasn't reproduced in all environments, but
a good way of reproducing this is putting a sleep in the generate_blocks
function

I really don't understand how it would ever help reproducing the bug you described. But let's see, trying now with: (EDIT: so, yeah, it still always passes for me even with the timeout) 
```patch
diff --git a/tests/test_framework/bitcoind.py b/tests/test_framework/bitcoind.py
index b25899b..f773a4c 100644
--- a/tests/test_framework/bitcoind.py
+++ b/tests/test_framework/bitcoind.py
@@ -106,6 +106,9 @@ class BitcoinD(TailableProc):
             else:
                 wait_for(lambda: len(self.rpc.getrawmempool()) >= wait_for_mempool)
 
+        import time
+        time.sleep(10)
+
         old_blockcount = self.rpc.getblockcount()
         addr = self.rpc.getnewaddress()
         self.rpc.generatetoaddress(numblocks, addr)
```

> In fact, the second spend was broadcasted one block before the first one,
and thus CPFPed one block before

Huh? No. We only CPFP transactions when we detect a new tip (and we broadcast the spends afterward). Hmm i think this could be the reason for your misunderstanding in the reproduction above.


-----

So, without digging more into it i can guess there could be a race with the tip update (as i mentioned in PM) but i'm very unconvinced by your description of the bug and the fact you are the only one experiencing it (it passes for me, Ed, and the CI).

src/daemon/bitcoind/poller.rs
danielabrozzoni: > This is in essence the same FIXME as the above:

No? My FIXME says that if we include in to_cpfp a transaction whose cpfp output was already spent, we fail to cpfp everything. Your comment isn't saying that?

darosior: Your FIXME is fixed by mine is what i was saying, but nevermind it's really not important :)

danielabrozzoni:
You were right: I thought the source of the race was the fact that the two spends were broadcastable at different block heights, but that wasn't it. In some cases, the spends would be CPFPed in different blocks, and this caused the test failure. Slowing down blocks should be enough to fix.

darosior:
This is very surprising that a race would constantly materialize in your tests but not everyone else's.

tests/test_misc.py
```diff
@@ -397,7 +398,12 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
         assert txid in cpfp_entry["depends"]
 
     # Now get to be able to broadcast the second Spend.
-    bitcoind.generate_block(CSV, wait_for_mempool=unvaults)
+    bitcoind.generate_block(1, wait_for_mempool=unvaults)
+    # We mine them slowly to avoid race conditions
+    for i in range(0, CSV - 1):
+        bitcoind.generate_block(1)
+        time.sleep(1)
```
darosior, line 405: Could you just get to `CSV - 1`, then mine `1` with `wait_for_mempool`?

darosior (commented):
If this is the cause then you'd have to do that for all the Spend tests? Or maybe we always mine up to the height - 1 then mine a single block in the Spend tests, as i suggested in the review comment.

tests/test_misc.py
danielabrozzoni: This `wait_for_mempool` doesn't work. You need to make sure to give revaultd to process height `CSV - 1` before mining the block `CSV`, otherwise you'll have the second spend broadcasted at height `CSV` AND cpfped in the same block.

darosior: No, Spends are broadcasted *after* we check for CPFP. Therefore you can't have a Spend that it CPFPed in the same block as it is broadcast (although i think it'd conceptually make sense).

danielabrozzoni:
You can.
You are at block 120. You mine CSV = 6 blocks. Revaultd catches 125 as the next tip. It tries to broadcast the first spend, it doesn't work. In the meanwhile block 126 gets mined, revaultd doesn't know yet, it tries to broadcast the second spend and it succeeds.
At this point revaultd notices a new block (126) arrived, and CPFPs the second spend.

I can post logs again, but that's what's happening.

darosior:
Yes, this is what i said. It's not in the same block as it's broadcast... Let me look into it.

darosior:
So i went through the test. As far as i can tell, it is correct. I have detailed it in this patch with annotations and making explicit its assumptions around where it's failing on your system:
```patch
index 60250e9..8c10dd7 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -433,7 +433,7 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
 
     # At this point, one of the first unvaults gets mined, but not the other. Feerate
     # spikes and makes us feebump: we'll create a CPFP tx spending the remaining unconfirmed
-    # Unvault from the first batch, and the 3 unvaults of the second spend.
+    # Unvault from the first batch, and the 2 unvaults of the second spend.
     revault_network.bitcoind_proxy.mocks["estimatesmartfee"] = {
         "feerate": 30 * 1_000 / COIN  # 30 is 6sats/vb above, should trigger CPFP
     }
@@ -452,8 +452,11 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
     for txid in unvaults:
         assert txid in cpfp_entry["depends"]
 
-    # Now get to be able to broadcast the second Spend.
-    bitcoind.generate_block(CSV, wait_for_mempool=unvaults)
+    # Now get to be able to broadcast the second Spend. Note that both the Spend
+    # transactions are only broadcastable at the last block, and therefore we can't
+    # get into a situation where they get broadcasted at block CSV-1 and CPFPed alone
+    # at block CSV.
+    bitcoind.generate_block(CSV, wait_for_mempool=unvaults + [cpfp_txid])
     man.wait_for_logs(
         [
             f"broadcasted Spend tx '{first_spend}'",
@@ -461,13 +464,24 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
         ]
     )
 
+    # Both spends are in the mempool!
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 2)
+
     # In the meantime, we attempt a third Spend.
     third_spend_psbt = revault_network.broadcast_unvaults_anyhow(
         thi_batch, priority=True
     )
     third_spend = third_spend_psbt.tx.hash
 
-    # If they don't get mined we'll CPFP all at once.
+    # Both spends and the Unvaults!
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 2 + len(third_unvaults))
+    # We have therefore 5 vaults in 'spending' and 5 in 'unvaulting'!
+    wait_for(lambda: len(man.rpc.listvaults(["spending"])["vaults"]) == 5)
+    wait_for(lambda: len(man.rpc.listvaults(["unvaulting"])["vaults"]) == 5)
+
+    # If they don't get mined we'll CPFP all at once. Generating a block triggers the
+    # check for CPFP, and 'should_cpfp' will return true for the 7 vaults we detailed
+    # above!
     to_be_cpfped = [first_spend, second_spend] + third_unvaults
     bitcoind.generate_blocks_censor(1, to_be_cpfped)
     man.wait_for_log(
```
Please point any anotation here you think isn't right.

I don't think there is anything wrong with the test, as hinted by the fact your failure isn't reproducible and my wannabe-demonstration using the above code annotations.

darosior:
So Daniela pointed out in private that actually one of the annotations above is wrong: 
> ```patch
> +    # [...] Note that both the Spend
> +    # transactions are only broadcastable at the last block, and therefore we can't
> +    # get into a situation where they get broadcasted at block CSV-1 and CPFPed alone
> +    # at block CSV.
> ```

They actually *can* get broadcasted at CSV-1 and trigger the CPFP logic at block CSV if revaultd polls twice. As detailed in the following patch: 
```patch
index 60250e9..f14d927 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -433,7 +433,7 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
 
     # At this point, one of the first unvaults gets mined, but not the other. Feerate
     # spikes and makes us feebump: we'll create a CPFP tx spending the remaining unconfirmed
-    # Unvault from the first batch, and the 3 unvaults of the second spend.
+    # Unvault from the first batch, and the 2 unvaults of the second spend.
     revault_network.bitcoind_proxy.mocks["estimatesmartfee"] = {
         "feerate": 30 * 1_000 / COIN  # 30 is 6sats/vb above, should trigger CPFP
     }
@@ -452,14 +452,24 @@ def test_batched_cpfp_transaction(revault_network, bitcoind):
     for txid in unvaults:
         assert txid in cpfp_entry["depends"]
 
-    # Now get to be able to broadcast the second Spend.
-    bitcoind.generate_block(CSV, wait_for_mempool=unvaults)
+    # Now get to be able to broadcast the second Spend. Note that both Spend transactions
+    # are only broadcastable at the last block, but revaultd might be lagging behind: it
+    # is possible it polls the tip at block CSV-1 and actually manage to broadcast a Spend
+    # since bitcoind is already at block CSV. It would then trigger CPFP for the Spend(s)
+    # alone when mining the last block, and fail the test later on when we assert that we
+    # CPFP the two spends and the unvaults at once.
+    # Therefore we make sure revaultd is sync'ed before we mine the last block.
+    height = man.rpc.getinfo()["blockheight"]
+    bitcoind.generate_block(CSV - 1, wait_for_mempool=unvaults + [cpfp_txid])
+    wait_for(lambda: man.rpc.getinfo()["blockheight"] == height + CSV - 1)
     man.wait_for_logs(
         [
             f"broadcasted Spend tx '{first_spend}'",
             f"broadcasted Spend tx '{second_spend}'",
         ]
     )
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 2)
 
     # In the meantime, we attempt a third Spend.
     third_spend_psbt = revault_network.broadcast_unvaults_anyhow(
```
@danielabrozzoni does this patch make sense to you? Does it solve the race you experienced?

darosior:
> Actually, I'm saying the same thing in #345 (comment) and #345 (comment) as well, but nevermind :)

image

darosior (commented):
ACK be8a914ab9312496f18ab20fea495f8898000362

PR #352, opened 2022-01-11
Refactor: use `listsinceblock` in...

danielabrozzoni:
...`sync_deposits`, `sync_unvaults`

At the moment `test_not_announceable_spend` timeouts often even in local environment, and I suspect the reason for this is that `sync_unvaults` is slower now if many vaults are being spent

Based on #344

### DIFF
--- a/src/bitcoind/interface.rs
+++ b/src/bitcoind/interface.rs
@@ -1,12 +1,8 @@
 use crate::config::BitcoindConfig;
 use crate::{bitcoind::BitcoindError, revaultd::BlockchainTip};
-use revault_tx::{
-    bitcoin::{
-        blockdata::constants::COIN_VALUE, consensus::encode, util::bip32::ChildNumber,
-        util::psbt::PartiallySignedTransaction as Psbt, Amount, BlockHash, OutPoint, Script,
-        Transaction, TxOut, Txid,
-    },
-    transactions::{DUST_LIMIT, UNVAULT_CPFP_VALUE},
+use revault_tx::bitcoin::{
+    consensus::encode, util::bip32::ChildNumber, util::psbt::PartiallySignedTransaction as Psbt,
+    Address, Amount, BlockHash, OutPoint, Script, Transaction, TxOut, Txid,
 };
 
 use std::{
@@ -23,10 +19,6 @@ use jsonrpc::{
     simple_http::{Error as HttpError, SimpleHttpTransport},
 };
 use serde_json::Value as Json;
-
-// The minimum deposit value according to revault_tx depends also on the unvault's
-// transaction fee. To have a one-value-fits-all, just take a 5% leeway.
-const MIN_DEPOSIT_VALUE: u64 = (DUST_LIMIT + UNVAULT_CPFP_VALUE) * 105 / 100;
 
 // If bitcoind takes more than 3 minutes to answer one of our queries, fail.
 const RPC_SOCKET_TIMEOUT: u64 = 180;
@@ -525,28 +517,6 @@ impl BitcoinD {
         self.import_fresh_descriptor(descriptor, UNVAULT_UTXOS_LABEL.to_string())
     }
 
-    pub fn list_unspent_deposits(
-        &self,
-        min_amount: Option<u64>,
-    ) -> Result<Vec<ListUnspentEntry>, BitcoindError> {
-        self.list_unspent(
-            &self.watchonly_client,
-            min_amount,
-            Some(DEPOSIT_UTXOS_LABEL),
-        )
-    }
-
-    pub fn list_unspent_unvaults(
-        &self,
-        min_amount: Option<u64>,
-    ) -> Result<Vec<ListUnspentEntry>, BitcoindError> {
-        self.list_unspent(
-            &self.watchonly_client,
-            min_amount,
-            Some(UNVAULT_UTXOS_LABEL),
-        )
-    }
-
     pub fn list_unspent_cpfp(&self) -> Result<Vec<ListUnspentEntry>, BitcoindError> {
         // For some weird reason, listunspent with the cpfp wallet doesn't return the label
         // (maybe because we only have one descriptor anyways?), so we pass `None` as a label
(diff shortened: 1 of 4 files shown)


darosior (commented):
Posting my incomplete review so we can discuss `listsinceblock` not returning outputs paying to us if they are not in the address book (which it interprets as change).

src/bitcoind/interface.rs
```diff
@@ -599,46 +572,126 @@ impl BitcoinD {
...
+                if let Some(confirmations) = confirmations {
+                    if confirmations >= min_conf as i32 {
+                        new_conf.insert(
+                            outpoint.clone(),
+                            UtxoInfo {
+                                txo: info.txo.clone(),
+                                is_confirmed: true,
+                            },
+                        );
+                    }
```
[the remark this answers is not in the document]
danielabrozzoni: Yeah I figured it wouldn't be too much of a problem if you go from `unconfirmed` to `unvaulting` (even tho you would have used some external software for signing all the txs...), but yeah, I never explicitly tested that

danielabrozzoni:
I might have found a bug while testing the GUI... Posting this comment to remind myself to check what happens if I have multiple transactions that pay the same address in the same block

src/bitcoind/interface.rs
```diff
@@ -647,28 +700,48 @@ impl BitcoinD {
...
+                for input in tx.input {
+                    if let Some(d) = unvault_utxos.get(&input.previous_output) {
+                        new_spent.insert(
+                            input.previous_output,
+                            UtxoInfo {
+                                txo: d.txo.clone(),
+                                is_confirmed,
+                            },
+                        );
+                    }
```
darosior, line 743: I think this might be the cause of the timeouts you mentioned. Imagine a Spend transaction with 588 outputs (as we do in the not announceable spend test) and 20 inputs (same test). This would loop 11_760 times.. (and this is combinatorial, this can quickly get much bigger).
I think an approach checking if all our utxos are still unspent using `gettxout`, and then using `listsinceblock` to detect new outputs paying to us (`receive`) would be reasonable.
Even for the deposits, i think using the `send` results from `listsinceblock` is a bit "clumsy" and an approach coupling `gettxout` for detecting the spent utxos and `listsinceblock` for detecting the received ones should be prefered.

src/bitcoind/interface.rs
```diff
@@ -599,46 +572,126 @@ impl BitcoinD {
...
+
+        // First, we check the unconfirmed deposit, to find out if they have enough
+        // confirmations now.
+        for (outpoint, info) in deposits_utxos {
+            if !info.is_confirmed {
+                let confirmations = self.get_unspent_tx_confirmations(&outpoint.txid)?;
+                if let Some(confirmations) = confirmations {
+                    if confirmations >= min_conf as i32 {
+                        new_conf.insert(
+                            outpoint.clone(),
```
darosior, line 617: `OutPoint` is `Copy`

src/bitcoind/interface.rs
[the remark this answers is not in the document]
darosior: I think it's fine

danielabrozzoni:
@darosior should be fixed now 🚀 If it's ok I'll squash

danielabrozzoni:
I just noticed that without using `send` I still have to refactor the code for the rescan to make sure to update the cache before the poller it's started, as I will notice as spent only utxos that were previously in the cache (same as it is now) :(

darosior:
Hard to say without looking at the rescan code but can't it use 2 polls? The first one adds the deposits to the cache/DB, the second one notices it's spent. And that's the same code path.

I do believe that's a still an improvement for rescan/"get back up" as the current code would not see a vault that was created and spent while the wallet isn't up (this can happen for instance for a manager in a deployment w/ a threshold, or a bypass). Oh, this actually makes me think this new vital feature introduced should be tested :D.

src/bitcoind/interface.rs
```diff
@@ -628,17 +628,24 @@ impl BitcoinD {
...
-                        utxo.outpoint,
-                        UtxoInfo {
-                            txo: utxo.txo,
-                            // All new utxos are marked as unconfirmed. This allows for a proper state
-                            // transition.
-                            is_confirmed: false,
-                        },
-                    );
+                    // Alright, I don't have this outpoint in the cache... But maybe I don't have
+                    // it because it's spent already! Let's check.
```
darosior, line 655: What is a scenario where this check is needed?

darosior (commented):
Yes, please go ahead and squash :)

src/bitcoind/interface.rs
danielabrozzoni: Let's keep in mind that `sync_deposits` could be run multiple times with the same `previous_tip`.
If between the `previous_tip` and the `current_tip` we receive and spend a deposit (let's call it D), at the first call of `sync_deposits` we will mark D as unconfirmed, at the second one as confirmed, at the third one as spent (the poller will proceed to remove it from the deposits cache), at the fourth one unconfirmed again.
You can reproduce a situation similar to this one removing the "already spent check" and running `test_reorged_deposit_status` - I believe what's happening in the test is slightly different, due to all the reorgs etc, but the main idea behind the bug should be the same

danielabrozzoni:
> Oh, this actually makes me think this new vital feature introduced should be tested :D.

Can I test this in the rescan PR or is it better to do it here?

darosior:
This feature is introduced in this PR, so i think it should be tested here.

src/bitcoind/interface.rs
darosior: I think this prevents syncing the history you missed (https://github.com/revault/revaultd/pull/352#issuecomment-1013754522). And that's the entire point of refactoring this code to use `listsinceblock`..

You are right that we are missing information here, Checking whether the new deposit does not correspond to a known spent vault should probably be done by the caller, that's bulky but probably fine as it would be a very edgy case (that i still fail to *precisely* model) that would be resolved by a new block.

> If between the previous_tip and the current_tip we receive and spend a deposit (let's call it D), at the first call of sync_deposits we will mark D as unconfirmed, at the second one as confirmed, at the third one as spent (the poller will here proceed to remove D from the deposits cache), at the fourth one unconfirmed again.

So we need to poll 4 times with the same `previous_tip` and `current_tip`? Therefore it must be the latest block. So you'd need a deposit to become confirmed, unvaulted and spent in the same block. That's not possible. Is there an edge case where this can happen? Maybe a deep reorg where the deposit shifts many blocks back?
Do you have a concrete example where this happens?

darosior:
So, discussing with Daniela about this on IRC we realized this actually needs to modify the rescan logic to use the same approach (get to know all events since the forked block, instead of "checking our state at the new chain tip").

On the top of my head, there are different approaches to this:
- Finding the common ancestor (between your fork and the new main chain) by repeatedly calling `getheader <tip> <1>` starting from the current tip and repeating on `previousblockhash` until `confirmations` isn't `-1` anymore
- Assuming there won't be re-orgs larger than 144 blocks (that's definitely reasonable to assume that if that happens Bitcoin's "finality" mechanism has failed and there is nothing we can do) and always rescan from this point

Regarding how to proceed once you spotted the fork points, i'll sleep on it but i'd be more inclined to do something "radical", i think.

src/bitcoind/interface.rs
```diff
@@ -599,46 +569,113 @@ impl BitcoinD {
...
+                        UtxoInfo {
+                            txo: info.txo.clone(),
+                            is_confirmed: true,
+                        },
+                    );
+                }
+            } else {
+                // We've seen it unspent and now it's not here anymore: it's spent.
+                new_spent.insert(*outpoint, info.clone());
+            }
```
darosior, line 646: Note so i don't forget: this actually introduces the possibility for a vault to go from `unconfirmed` to `unvaulting` immediately.

I'm not sure how we should approach to patch this yet.

danielabrozzoni:
Closing in favor of #366 :D

PR #355, opened 2022-01-20
Add trait DaemonCommands

edouardparis:


src/commands/mod.rs
```diff
@@ -225,26 +227,26 @@ impl DaemonControl {
...
-    pub fn get_deposit_address(&self) -> Address {
-        self.revaultd.read().unwrap().deposit_address()
+    fn get_deposit_address(&self) -> Address {
+        self.revaultd().deposit_address()
     }
 
     // Internal only, used for testing
-    pub(crate) fn get_deposit_address_at(&self, index: bip32::ChildNumber) -> Address {
-        self.revaultd.read().unwrap().vault_address(index)
+    fn get_deposit_address_at(&self, index: bip32::ChildNumber) -> Address {
```
darosior, line 248: I'd prefer to not expose this

src/commands/mod.rs
```diff
@@ -187,10 +187,12 @@ macro_rules! manager_only {
     };
 }
 
-impl DaemonControl {
+pub trait DaemonCommands {
+    fn revaultd(&self) -> RwLockReadGuard<RevaultD>;
+    fn bitcoind(&self) -> &dyn BitcoindThread;
```
darosior, line 192: nit: why `dyn` and not `impl`?

darosior (commented):
So i see the benefit (which you mentioned privately to me) to get rid of the circular dep between `lib` and `commands`, but i'm wary that we'd get the `trait` "craze" by trying to introduce trait in a lot of places. This increases technical debt as it tends to transform any small rework into a large refactoring.
Also, i think we should first leverage traits to generalize the internals, and then get to what depends on them. For instance, i'd like to generalize the DB (at least make it safer, maybe make it have its own thread). Maybe `RevaultD` could be as well which could help with unit testing to mock instead of creating dummy datadirs.

PR #357, opened 2022-01-24
Batch the exchange of signatures with the watchtowers

darosior:
As per https://github.com/revault/practical-revault/pull/115.
This greatly simplifies the implementation, too!


darosior (commented):
self-ACK 003a308811f387e0737e28ff04ba542b533de6da

PR #359, opened 2022-01-28
Add Deserialize macro to commands response

edouardparis:
Clients importing the revaultd lib and consuming
the jsonrpc API need to deserialize the commands
responses.

This commit introduces new deserialization helpers
and adds them to commands structs fields.

### DIFF
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1446,7 +1446,7 @@ pub struct ListOnchainTxEntry {
 }
 
 /// Status of a Spend transaction
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub enum ListSpendStatus {
     NonFinal,
@@ -1455,7 +1455,7 @@ pub enum ListSpendStatus {
 }
 
 /// Information about a Spend transaction
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ListSpendEntry {
     pub deposit_outpoints: Vec<OutPoint>,
     pub psbt: SpendTransaction,
@@ -1472,7 +1472,7 @@ pub struct ServersStatuses {
 }
 
 /// The type of an accounting event.
-#[derive(Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum HistoryEventKind {
     #[serde(rename = "cancel")]
     Cancel,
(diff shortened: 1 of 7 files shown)


darosior:
Mobile review ACK. Two nits:

- Why prefixing the deser routines with `der` (as opposed to `deser` for instance)?

- Can you name the amount deser routine as `deser_sats` instead of `deser_amount` to mitigate future misuses?

darosior:
wtf is going on with the Python linter

darosior:
Cool thanks for rebasing on top of the linter fix

src/commands/mod.rs
```diff
@@ -1395,6 +1395,12 @@ pub struct ListVaultsEntry {
     pub moved_at: Option<u32>,
 }
 
+impl ListVaultsEntry {
+    pub fn outpoint(&self) -> OutPoint {
+        OutPoint::new(self.txid, self.vout)
+    }
+}
```
darosior, line 1399: Nothing is missing for a client of the API to do that on their own?

I mean i'm not against including it, it doesn't hurt. But then one probably wants that for something else as well, and another thing, etc.. While all the fields are `pub` and they can just use them.

src/commands/mod.rs
```diff
@@ -226,14 +225,12 @@ impl DaemonControl {
     /// List the current vaults, optionally filtered by status and/or deposit outpoints.
     pub fn list_vaults(
         &self,
-        statuses: Option<Vec<VaultStatus>>,
-        deposit_outpoints: Option<Vec<OutPoint>>,
-    ) -> ListVaultsResult {
+        statuses: Option<&[VaultStatus]>,
```
darosior, line 228: Neat

src/commands/mod.rs
```diff
@@ -860,7 +860,7 @@ impl DaemonControl {
     pub fn get_spend_tx(
         &self,
         outpoints: &[OutPoint],
-        destinations: BTreeMap<Address, u64>,
+        destinations: &BTreeMap<Address, u64>,
```
darosior, line 863: Hmmm i thought we consumed it. But the `collect()` in
```rust
        let txos: Vec<SpendTxOut> = destinations
            .into_iter()
            .map(|(addr, value)| {
                let script_pubkey = addr.script_pubkey();
                SpendTxOut::new(TxOut {
                    value: *value,
                    script_pubkey,
                })
            })
            .collect();
```
seems to have the effect of a clone.

I'd prefer if we moved it but it's not important.

darosior (commented):
ACK fded40da8515b59a7f9a210b4c08f7b2cea476c3

Have a look at the comments but i'm happy to merge it as is.

src/commands/mod.rs
edouardparis: Ok I will remove it

src/commands/mod.rs
edouardparis: `script_pubkey()` method need only a reference and `value` is a u64: no need to clone the btreemap values and I think `into_iter` does not clone it (I will change for iter()).
I think it is good practice that if your method does not need to own an object to pass only a ref to it

darosior: Yes it definitely is. As mentioned in my comment, my mistake i figured it had to be moved :)

darosior (commented):
utACK 24eef7d5a4189fb7b5c2252ceaebedb630c6f936

darosior:
I didn't really question the rationale here. Can you remind me why would a user of the library need `Deserialize` necessarily?

PR #363, opened 2022-02-04
ci: update files so they use black-22

danielabrozzoni:


darosior:
We should probably pin `black`, too.

darosior:
re-utACK [b775a7c](https://github.com/revault/revaultd/pull/363/commits/b775a7c95a2a2957159c8eae30e5e9b942323716)

PR #361, opened 2022-02-01
Test that the coordinator broadcasts spend txs

danielabrozzoni:
Needs https://github.com/revault/coordinatord/pull/46

PR #364, opened 2022-02-07
Compute checksums for wallet descriptors without bitcoind

### LINKED ISSUE #322: Don't use `bitcoind`'s RPC for the watchonly wallet descriptors checksums
For the watchonly wallet, we import `addr()` descriptors to `bitcoind` since it does not handle Miniscript yet. A descriptor must have a checksum to be imported to bitcoind. Back when i implemented this rust-miniscript did not have support for descriptor checksums. I used the hack of calling `getdescriptorinfo` before importing a descriptor to gather the checksum *for each and every descriptor we import* (hundreds on first startup).
Now that rust-miniscript does support them it's an unnecessary hack and performance hit. Replace it with just using the checksums provided by rust-miniscript.

JSwambo:
The desc_checksum function and poly_mod were taken from rust-miniscript: https://github.com/rust-bitcoin/rust-miniscript/blob/master/src/descriptor/checksum.rs
Hopefully the desc_checksum function will be publicly exposed in the future and we can remove it.

Fixes #322


JSwambo:
When I test locally I don't get the same failure with `test_spend.py::test_large_spends`. Not sure why this is failing

darosior:
Yeah, spurious failure fixed by #361 . Rebasing on master should fix it.

darosior (commented):
The cosigner submodule change looks unrelated?..

PR #367, opened 2022-02-10
Poller reorg cleanup

darosior:
This is a followup to #366 which cleans up the reorg logic, trying to make the code paths easier to follow. It then removes all RPC calls from within the DB transaction used to rewind the state (:tada:). Finally it adds a more thorough test for the timestamp fields as requested in #366 and includes a couple drive-by cleanups.


darosior:
Rebased on #366

darosior:
self-ACK 69bfc3fee951966831bf73b6dea2fbef921fbbfb

It's blocking and i've been building on it for some time now without any issue.

PR #376, opened 2022-02-23
Test sigs fetching for secure vaults

edouardparis:
Test when a vault is secured and a user starts
signing the unvault tx, every other participants
retrieve and store the signature.

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -250,6 +250,45 @@ def test_sigfetcher_coordinator_dead(revault_network, bitcoind):
 
 
 @pytest.mark.skipif(not POSTGRES_IS_SETUP, reason="Needs Postgres for servers db")
+def test_sigfetcher_secured_vaults(revault_network, bitcoind):
+    """
+    Test that unvault sigs are retrieved and stored for secured vault even if the
+    daemon user did not start activating it or is not a stakeholder.
+    """
+    rn = revault_network
+    rn.deploy(
+        3,
+        1,
+        csv=1,
+        with_cosigs=False,
+        with_watchtowers=False,
+    )
+    vault = rn.fund(1)
+    rn.secure_vault(vault)
+    stks = rn.stks()
+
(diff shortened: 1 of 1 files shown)


tests/test_misc.py
```diff
@@ -249,6 +249,56 @@ def test_sigfetcher_coordinator_dead(revault_network, bitcoind):
         stk.wait_for_secured_vaults([deposit])
 
 
+# test that unvault sigs are retrieved and stored for secured vault even if the
+# daemon user did not start activating it or is not a stakeholder.
```
darosior, line 253: nit: can you make it a docstring? That's how we do it for the other tests
```python
def f(a, b):
    "Adds a and b"""
    return a + b
```

tests/test_misc.py
```diff
@@ -249,6 +249,56 @@ def test_sigfetcher_coordinator_dead(revault_network, bitcoind):
...
+            [
+                "Syncing Unvault signature",
+            ]
+        )
+        wait_for(
+            lambda: man.rpc.listpresignedtransactions([outpoint])[
+                "presigned_transactions"
+            ][0]["unvault"]["psbt"]
+            == unvault_psbt
+        )
```
darosior, line 299: You can merge those loops using `rn.participants()` :)

tests/test_misc.py
```diff
@@ -249,6 +249,45 @@ def test_sigfetcher_coordinator_dead(revault_network, bitcoind):
...
+    for stk in rn.participants():
+        stk.wait_for_logs(
+            [
+                "Syncing Unvault signature",
+            ]
+        )
+        wait_for(
+            lambda: stk.rpc.listpresignedtransactions([outpoint])[
+                "presigned_transactions"
+            ][0]["unvault"]["psbt"]
```
darosior, line 286: This would silently conflict with #377. But let's merge this first and i'll amend it in #377.

darosior (commented):
utACK 6b5af941029a878d18db560fe6dbcea0061dc26f

Issue #324, opened 2021-12-03
Check bitcoind version at startup


darosior:
`bitcoind`'s releases are versioned. New releases bring new features. `revaultd` depends on the newest features of `bitcoind`. Checking the version at startup is for sanity checking that the features we depend on are available on the `bitcoind` we are configured to talk to.

gourab337:
@darosior I added a pr some days back: https://github.com/revault/revaultd/pull/378

darosior:
@gourab337 please note that a PR needs to pass CI before being considered. Also, your PR does not seem to fix this issue in its current form.

PR #378, opened 2022-03-20
Check bitcoind version at startup #324

gourab337:
The function bitcoind_version_check() is unused.

### DIFF
--- a/src/bitcoind/mod.rs
+++ b/src/bitcoind/mod.rs
@@ -144,6 +144,53 @@ pub fn start_bitcoind(revaultd: &mut RevaultD) -> Result<BitcoinD, BitcoindError
     Ok(bitcoind)
 }
 
+/// Check bitcoind version at startup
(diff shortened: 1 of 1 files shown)


meryacine:
Hello @gourab337. I submitted another [PR](#385) solving the same issue while not noticing this one, Sorry :sweat_smile:.
You may want to use the RPC interface to query bitcoind for its version instead of running a process. The PR I linked above does so. But it didn't pass the CI. I found a mistake in it that is fixed in this [branch](https://github.com/meryacine/revaultd/tree/bitcoind_at_startup) (it was about the version number being an `u64` and not a string), also added a simple test for the `getnetworkinfo` interface. Please feel free to use that code if it could help with this PR :blush:.

darosior:
Let's leave @gourab337 a couple more days to answer, after that feel free to grab this @meryacine.

Issue #45, opened 2020-12-07
Add participant kind in `getinfo`


darosior:
I think @edouardparis will need to know if the `revaultd` is configured for a stakeholder or a manager without actually parsing the config ?

edouardparis:
revault-gui need to parse the config to get the datadir and find the socket path, It can parse the `user role` in the config

darosior:
Hmm right, but i still think we should do it for applications that don't want to be parsing the config

edouardparis:
ACK :+1:

darosior:
Similarly to #174 this is about adding a new field in `getinfo`.

The key could be `participant_type` and the value should be the "kind" of participants, `stakeholder`, `stakeholdermanager`, or `manager`.
Useful skills for solving this issues:
- Knowing what these kinds correspond to
- Knowledge about the config parsing, to see where (and on what basis) we set whether a participant is a stk, stkman or man
- Being familiar with the `RevaultD` "main state data structure" to be able to forward this information to the JSONRPC interface

darosior:
Done in #387

PR #379, opened 2022-03-23
Added participant_type to get_info.

Zshan0:
Hey, I added `participant_type` to `get_info` as requested in #45.

Let me know if there are any changes to be made.


src/commands/mod.rs
```diff
@@ -1430,6 +1440,7 @@ pub struct GetInfoResult {
     pub vaults: usize,
     pub managers_threshold: usize,
     pub descriptors: GetInfoDescriptors,
+    pub participant_type: String,
```
edouardparis, line 1443: Thank you for the PR.

Please use [`UserRole`](https://github.com/revault/revaultd/blob/master/src/utils.rs#L28), you may need to implement `fmt::Display` to it and add ` #[serde(serialize_with = "ser_to_string", deserialize_with = "deser_from_str")]` to the `GetInfoResult` field

Zshan0: Hey, I was looking through the codebase and I found that `UserRole` has been declared under `#[cfg(test)]` which means that it can't be used outside tests, I confirmed it by looking through the pre-existing uses of `UserRole` which have all been used in tests only.

Maybe I'm missing something but could you direct me towards a possible solution?

edouardparis: You are right, I think it is better to move `UserRole` to revaultd.rs and outside of the #[cfg(test)].
Best case would be to have in this PR:
1. a commit to move UserRole to revaultd.rs and modify every tests using it
2. a commit to introduce `participant_type` to `getinfo` command

Sorry for the additional work :)

Zshan0: Okay great, I'll start working on that then :D

Zshan0:
I've made the requested changes.

One thing that I noticed is that the string form of `UserRole` is the string that was requested is `stakeholdermanager` but the enum `ManagerStakeholder`, please let me know if I should change one of them to the other to avoid confusion

src/revaultd.rs
```diff
@@ -276,6 +276,41 @@ pub struct BlockchainTip {
     pub hash: BlockHash,
 }
 
+#[derive(Debug, Clone)]
+pub enum UserRole {
+    Stakeholder,
+    Manager,
+    ManagerStakeholder,
```
edouardparis, line 283: Ok to change for StakeholderManager (It is the convention we use in other repositories like `aquarium`)

src/revaultd.rs
```diff
@@ -276,6 +276,41 @@ pub struct BlockchainTip {
...
+}
+
+impl FromStr for UserRole {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "stakeholder" => Ok(Self::Stakeholder),
+            "stakeholdermanager" => Ok(Self::Manager),
```
edouardparis, line 292:
```suggestion
            "stakeholdermanager" => Ok(Self::StakeholderManager),
```

src/commands/mod.rs
```diff
@@ -268,6 +268,15 @@ impl DaemonControl {
             })
             .count();
 
+        assert!(revaultd.is_stakeholder() || revaultd.is_manager());
+        let participant_type = if revaultd.is_manager() && revaultd.is_stakeholder() {
+            UserRole::ManagerStakeholder
+        } else if revaultd.is_manager() {
+            UserRole::ManagerStakeholder
```
edouardparis, line 275:
```suggestion
            UserRole::Manager
```

src/revaultd.rs
```diff
@@ -1,4 +1,4 @@
-use crate::{
+    use crate::{
```
darosior, line 1: Looks spurious.

darosior (commented):
Cool, thanks for tackling this! Looks good to me but please [squash](https://git-scm.com/docs/git-rebase#_interactive_mode) your commits and format your code to make the linter happy.

Zshan0:
Since there is a change of `StakeholderManager` from `ManagerStakeholder`, should it have its own commit? There is an interleaving between the commits that are required in the same squash, which can't be done I think. So what do you propose the final list of squashed commits be?

darosior:
> Since there is a change of StakeholderManager from ManagerStakeholder, should it have its own commit?

I think so, yes.

> So what do you propose the final list of squashed commits be?

I think the best one would be the refactoring first, then the new feature.

----

Also -and i forgot this in my previous comment- any new feature needs tests. Given it's a small thing here, no need for it to be too elaborated. A simple addition to the sanity check for this RPC command in the functional tests would do:
https://github.com/revault/revaultd/blob/d014e662a5b3ace996362bfd2ba63fd1b6fad8ea/tests/test_rpc.py#L23-L33

darosior:
Needs rebase now that #384 is merged. Thanks for adding a test. This still needs commits to be squashed:  a6ac762d8501975248ee2465693457dab8a3ea2d and 9a5dbf703f4824cbf6425ccdc7198938eba10005 should be merged together.

Zshan0:
This PR can be closed since there is a more updated one with commits arranged as required

darosior:
Please not for next time to not duplicate pull requests, you can just force-push to this one.

PR #380, opened 2022-03-23
Add daemonize feature

edouardparis:
In order to make daemonize-simple optional


darosior (commented):
I'm not opposed to make daemonize-simple optional, but what is the rationale?

Note i think we could eventually have it in-tree too, since it's pretty simple and we could drop the Windows support.

PR #384, opened 2022-03-29
Pinning black to a new version.

Zshan0:
Pinning black to `22.3.0` as mentioned in https://github.com/psf/black/issues/2964


darosior (commented):
utACK 11e383a020b74b78ebd2ef18ca831e8eb0f7011f

Thanks for taking care of that annoyance.

Issue #300, opened 2021-10-27
Unit test `onchain_txs`


darosior:
#294 made it easier to write some unit tests for `control.rs`'s `onchain_txs` by mocking the `bitcoind` thread interface.

PR #386, opened 2022-04-01
Add cpfp and change index getonchaintransactions

edouardparis:
Raw hex transactions give not enough information about
the UTXO recipients. We introduce in this commit two more fields
to the list items: change_index and cpfp_index, respectively
the indexes of the change output and cpfp output in a transaction.

A cpfp output is an UTXO with a script pubkey derived from the CPFP descriptor.
A change output is an UTXO with a script pubkey derived from the Deposit
descriptor.

It adds a test (in relation to #300)

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -228,12 +228,14 @@ network (hence they may be unconfirmed). Will error if any of the vaults is unkn
 
 #### Wallet tx
 
-| Field         | Type          | Description                                                                   |
-| ------------- | ------------- | ----------------------------------------------------------------------------  |
-| `blockheight` | int or `null` | Height of the block containing the transaction, `null` if unconfirmed         |
-| `blocktime`   | int or `null` | Timestamp of the block containing the transaction, `null` if unconfirmed      |
-| `hex`         | string        | Hexadecimal of the network-serialized transaction                             |
-| `received_at` | int           | Transaction reception date as the number of seconds since UNIX epoch          |
+| Field          | Type          | Description                                                                  |
+| -------------- | ------------- | ---------------------------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


darosior:
Needs rebase :)

darosior (commented):
The implementation looks good to me, and thanks for adding a unit test to the not-unit-tested `listonchaintransactions`.

However, i don't think it is a good idea to expose a boutique change detection in `listonchaintransactions`.

Firstly, one can't find all the change outputs of a transaction with a reasonable degree of certitude short of checking all outputs' scriptPubKeys against your deposit descriptor up to the gap limit. It is obviously not reasonable. Short of doing that here we:
- Assume a change output will always reuse the same deposit address
- Only consider the last detected change output (because we reuse the hack from `listspendtxs` which is specific to the behaviour of `getspendtx`)

This will necessarily give **wrong** information to the user. Absence of such an information is obviously better.

Secondly, onchain transactions are necessarily past transactions. In order to detect change outputs from past transactions, we can just check if the txid of the given transaction is also the txid of a later deposit. It's expensive, but actually robust.
Good thing is: `gethistory` is doing just that. If one wants to know the amount that went back in for a given spending, they can just use that.

Lastly, i don't think that finding change outputs in historical transactions is a reasonable feature. It is not natural to check for change outputs without context, it's a limitation of how Bitcoin works. i think that an application relying on this should think twice about whatever they want to achieve as it sounds a lot like an anti-feature. (This holds for adding change outputs to the vault history view in the GUI, but really also for any other application that would need this).

For all these reasons, Concept NACK 535ab9b0b10854521cd22dfaaf7d681da72ce8e2.

Issue #95, opened 2021-02-22
Retry on connection error to bitcoind 


darosior:
If we fail to communicate with bitcoind, we'd currently crash. Instead it would be nice to retry until a timeout (say 60s as default for actual users?) which would be configurable so we can fail early in functional tests.

A good place to implement it would be in
https://github.com/re-vault/revaultd/blob/1f6b2d9b640aaf8db1fd8c9c4b1ca2000d2984b1/src/daemon/bitcoind/interface.rs#L70-L83

darosior:
The retry logic is [now implemented](https://github.com/revault/revaultd/blob/24401867e034a6ad3403ab367efad0f2183aa4cb/src/daemon/bitcoind/interface.rs#L84) but error handling could be improved.

PR #388, opened 2022-04-12
Refac bitcoind request retry after startup

### LINKED ISSUE #372: DaemonHandle should return error from `fn start()` if bitcoind config is not correct
My bitcoind config was not correct, yet revaultd still run with it:
```
❯ revaultd --config revaultd_regtest.toml 
[1645115202][revaultd::revaultd][WARN] CPFP key not found, consider creating a cpfp_secret file in the datadir. Automated CPFP won't be available.
[1645115202][revaultd][INFO] Using Noise static public key: '14ca9e4d387bccf35746e0407daaacc6b28a4f8445ef5a5158894db983e24070'
[1645115202][revaultd][INFO] Setting up database
[1645115202][revaultd][INFO] Setting up bitcoind connection
[1645115202][revaultd::bitcoind::interface][ERROR] Transport error when talking to bitcoind: 'Couldn't connect to host: Connection refused (os error 111)'
[1645115203][revaultd::bitcoind::interface][ERROR] Transport error when talking to bitcoind: 'Couldn't connect to host: Connection refused (os error 111)'
[1645115204][revaultd::bitcoind::interface][ERROR] Transport error when talking to bitcoind: 'Couldn't connect to host: Connection refused (os error 111)'
[1645115205][revaultd::bitcoind::interface][ERROR] Transport error when talking to bitcoind: 'Couldn't connect to host: Connection refused (os error 111)'
[1645115206][revaultd::bitcoind::interface][ERROR] Transport error when talking to bitcoind: 'Couldn't connect to host: Connection refused (os error 111)'
[1645115207][revaultd::bitcoind::interface][ERROR] Transport error when talking to bitcoind: 'Couldn't connect to host: Connection refused (os error 111)'
^C


```

edouardparis:
Code can not prevent Connection and I/O Errors to happen, in that case
the bitcoind client can retry multiple time its request in the given limit.
This commit implements the try limit and make the request try again
in case of i/o or timeout errors.

In order to fail faster at startup the try limit is set after the
bitcoind sanity checks.

close #372
closes https://github.com/revault/revaultd/issues/95

### DIFF
--- a/src/bitcoind/interface.rs
+++ b/src/bitcoind/interface.rs
@@ -5,13 +5,7 @@ use revault_tx::bitcoin::{
     Address, Amount, BlockHash, OutPoint, Script, Transaction, TxOut, Txid,
 };
 
-use std::{
-    any::Any,
-    collections::HashMap,
-    fs,
-    str::FromStr,
-    time::{Duration, Instant},
-};
+use std::{collections::HashMap, fs, str::FromStr, time::Duration};
 
 use jsonrpc::{
     arg,
@@ -34,6 +28,9 @@ pub struct BitcoinD {
     node_client: Client,
     watchonly_client: Client,
     cpfp_client: Client,
+
+    /// How many times the client will try again to send a request to bitcoind upon failure
+    retries: usize,
 }
 
 macro_rules! params {
@@ -89,64 +86,51 @@ impl BitcoinD {
             node_client,
             watchonly_client,
             cpfp_client,
+            retries: 0,
         })
     }
 
-    // Reasonably try to be robust to possible spurious communication error.
-    fn handle_error(&self, e: jsonrpc::Error, start: Instant) -> Result<(), BitcoindError> {
-        let now = Instant::now();
+    /// Set the retry limit (number of times we'll retry a request to bitcoind upon specific failures).
+    pub fn with_retry_limit(mut self, retry_limit: usize) -> Self {
+        self.retries = retry_limit;
+        self
+    }
 
-        match e {
-            jsonrpc::Error::Transport(ref err) => {
-                log::error!("Transport error when talking to bitcoind: '{}'", err);
-
-                // This is *always* a simple_http::Error. Rule out the error that can
-                // not occur after startup (ie if we encounter them it must be startup
-                // and we better be failing quickly).
-                let any_err = err as &dyn Any;
-                if let Some(http_err) = any_err.downcast_ref::<HttpError>() {
-                    match http_err {
-                        HttpError::InvalidUrl { .. } => return Err(BitcoindError::Server(e)),
-                        // FIXME: allow it to be unreachable for a handful of seconds,
-                        // but not at startup!
-                        HttpError::SocketError(_) => return Err(BitcoindError::Server(e)),
-                        HttpError::HttpParseError => {
-                            // Weird. Try again once, just in case.
-                            if now.duration_since(start) > Duration::from_secs(1) {
-                                return Err(BitcoindError::Server(e));
+    /// Wrapper to retry a request sent to bitcoind upon IO failure
+    /// according to the configured number of retries.
+    fn retry<T, R: Fn() -> Result<T, BitcoindError>>(
+        &self,
+        request: R,
+    ) -> Result<T, BitcoindError> {
+        let mut error: Option<BitcoindError> = None;
+        for i in 0..self.retries + 1 {
+            match request() {
+                Ok(res) => return Ok(res),
+                Err(e) => {
+                    if e.is_warming_up() {
+                        error = Some(e)
+                    } else if let BitcoindError::Server(jsonrpc::Error::Transport(ref err)) = e {
+                        match err.downcast_ref::<HttpError>() {
+                            Some(HttpError::Timeout)
+                            | Some(HttpError::SocketError(_))
+                            | Some(HttpError::HttpErrorCode(503)) => {
+                                std::thread::sleep(Duration::from_secs(1));
+                                log::debug!("Retrying RPC request to bitcoind: attempt #{}", i);
+                                error = Some(e);
                             }
-                            std::thread::sleep(Duration::from_secs(1));
+                            _ => return Err(e),
                         }
-                        _ => {}
+                    } else {
+                        return Err(e);
                     }
                 }
-
-                // This one *may* happen. For a number of reasons, the obvious one may
-                // be the RPC work queue being exceeded. In this case, and since we'll
-                // usually fail if we err try again for a reasonable amount of time.
-                if now.duration_since(start) > Duration::from_secs(45) {
-                    return Err(BitcoindError::Server(e));
-                }
-                std::thread::sleep(Duration::from_secs(1));
-                log::debug!("Retrying RPC request to bitcoind.");
             }
-            jsonrpc::Error::Json(ref err) => {
-                // Weird. A JSON serialization error? Just try again but
-                // fail fast anyways as it should not happen.
-                log::error!(
-                    "JSON serialization error when talking to bitcoind: '{}'",
-                    err
-                );
-                if now.duration_since(start) > Duration::from_secs(1) {
-                    return Err(BitcoindError::Server(e));
-                }
-                std::thread::sleep(Duration::from_millis(500));
-                log::debug!("Retrying RPC request to bitcoind.");
-            }
-            _ => return Err(BitcoindError::Server(e)),
-        };
+        }
 
-        Ok(())
+        Err(BitcoindError::Custom(format!(
+            "Retry limit reached: {:?}",
+            error
+        )))
     }
 
     fn make_request<'a, 'b>(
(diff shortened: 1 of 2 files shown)


src/bitcoind/interface.rs
```diff
@@ -89,64 +86,46 @@ impl BitcoinD {
...
+        request: R,
+    ) -> Result<T, BitcoindError> {
+        let mut error: Option<BitcoindError> = None;
+        for i in 0..self.try_limit {
+            match request() {
+                Ok(res) => return Ok(res),
+                Err(e) => match e {
+                    BitcoindError::Server(jsonrpc::Error::Transport(ref err)) => {
+                        match err.downcast_ref::<HttpError>() {
+                            Some(HttpError::Timeout) | Some(HttpError::SocketError(_)) => {
```
edouardparis, line 112: Rules for retry are now stricter: only timeout and i/o errors

src/bitcoind/interface.rs
```diff
@@ -34,6 +28,9 @@ pub struct BitcoinD {
     node_client: Client,
     watchonly_client: Client,
     cpfp_client: Client,
+
+    /// The number of try a client is allowed to do for a request.
+    try_limit: usize,
```
darosior, line 33: Nit: i think it's more intuitive for it to be the number of retries?
```suggestion
    /// How many times we'll try again to send a request to bitcoind upon failure
    retries: usize,
```

src/bitcoind/interface.rs
```diff
@@ -89,64 +86,46 @@ impl BitcoinD {
             node_client,
             watchonly_client,
             cpfp_client,
+            try_limit: 1,
```
darosior, line 89: This should be documented above that it will set the number of retries to 0 and the caller should use the introduced API to set it.

src/bitcoind/interface.rs
```diff
@@ -89,64 +86,46 @@ impl BitcoinD {
...
-                            if now.duration_since(start) > Duration::from_secs(1) {
-                                return Err(BitcoindError::Server(e));
+    /// Set the try limit field limiting the number of try a client is allowed to do for a request.
+    pub fn with_try_limit(mut self, retry_limit: usize) -> Self {
+        self.try_limit = retry_limit;
+        self
+    }
+
+    /// Code can not prevent Connection and I/O Errors to happen, in that case
+    /// the client can retry multiple time its request in the given limit.
```
darosior, line 100: I find
> Code can not prevent Connection and I/O Errors to happen

confusing. How about:
```suggestion
    /// Wrapper to retry a request sent to bitcoind upon IO failure according to the configured number of retries.
```

src/bitcoind/interface.rs
darosior: Are you sure that if the RPC work queue is exceeded it's going to be a timeout? I think it would be a 500 but i didn't [double-check](https://github.com/bitcoin/bitcoin/blob/623745ca74cf3f54b474dac106f5802b7929503f/src/rpc/protocol.h#L18).

I otherwise agree that my hacks for errors that shouldn't happen are unnecessary. :)

src/bitcoind/interface.rs
```diff
@@ -89,64 +86,46 @@ impl BitcoinD {
...
+    ) -> Result<T, BitcoindError> {
+        let mut error: Option<BitcoindError> = None;
+        for i in 0..self.try_limit {
+            match request() {
+                Ok(res) => return Ok(res),
+                Err(e) => match e {
+                    BitcoindError::Server(jsonrpc::Error::Transport(ref err)) => {
+                        match err.downcast_ref::<HttpError>() {
+                            Some(HttpError::Timeout) | Some(HttpError::SocketError(_)) => {
+                                std::thread::sleep(Duration::from_millis(500));
```
darosior, line 113: Any rationale for reducing this?

src/bitcoind/mod.rs
```diff
@@ -23,6 +23,10 @@ use jsonrpc::{
     simple_http,
 };
 
+/// Number of try the client is allowed to do in case of timeout and i/o error
+/// while communicating with the bitcoin daemon.
+const BITCOIND_TRY_LIMIT: usize = 6;
```
darosior, line 28: That a huge reduction in the number of retries. What is the rationale here? We would previously support spurious failures of `bitcoind`  for up to 1min. After this patch we'd fail after `6 * .5 = 3` seconds.

darosior (commented):
Cool, it's much cleaner than my previous approach! I assume you've tested it with the GUI?

src/bitcoind/interface.rs
edouardparis: https://github.com/bitcoin/bitcoin/blob/40a9037a1b5d990637d7f5009fc0c39628ed2c05/doc/release-notes/release-notes-22.0.md#rpc
You are right we should also retry upon HTTP_SERVICE_UNAVAILABLE

darosior:
Updated the OP as it solves https://github.com/revault/revaultd/issues/95 too.

src/bitcoind/interface.rs
```diff
@@ -89,64 +86,48 @@ impl BitcoinD {
...
-                    match http_err {
-                        HttpError::InvalidUrl { .. } => return Err(BitcoindError::Server(e)),
-                        // FIXME: allow it to be unreachable for a handful of seconds,
-                        // but not at startup!
-                        HttpError::SocketError(_) => return Err(BitcoindError::Server(e)),
-                        HttpError::HttpParseError => {
-                            // Weird. Try again once, just in case.
-                            if now.duration_since(start) > Duration::from_secs(1) {
-                                return Err(BitcoindError::Server(e));
+    /// Set the retry limit field limiting the number of retry a client is allowed to do for a request.
```
darosior, line 93: nit
```suggestion
    /// Set the retry limit (number of times we'll retry a request to bitcoind upon specific failures).
```

src/bitcoind/interface.rs
```diff
@@ -89,64 +86,48 @@ impl BitcoinD {
...
+        let mut error: Option<BitcoindError> = None;
+        for i in 0..self.retries + 1 {
+            match request() {
+                Ok(res) => return Ok(res),
+                Err(e) => match e {
+                    BitcoindError::Server(jsonrpc::Error::Transport(ref err)) => {
+                        match err.downcast_ref::<HttpError>() {
+                            Some(HttpError::Timeout)
+                            | Some(HttpError::SocketError(_))
+                            | Some(HttpError::HttpErrorCode(503)) => {
```
darosior, line 114: Hmm. I think we should retry on warmup (for instance if bitcoind was restarted from under us). It's RPC error `-28`. Is there anything else we'd like to cover? Maybe an http `500`? Not sure, it was used before 22.0 for limit exceeded but we don't support previous versions anyways. Looks like it's used to send JSONRPC errors too. Apart from warmup i think we want to let JSONRPC error go through right?

src/bitcoind/mod.rs
```diff
@@ -23,6 +23,11 @@ use jsonrpc::{
     simple_http,
 };
 
+/// Number of retry the client is allowed to do in case of timeout and i/o error
+/// while communicating with the bitcoin daemon.
+/// A retry happens every 1 second, after one minute the client stop trying.
```
darosior, line 28: supernit
```suggestion
/// Number of retries in case of timeout or i/o error
/// while communicating with the bitcoin daemon.
/// A retry happens every 1 second, this makes us give up after one minute.
```

darosior (commented):
I think we can still handle more failure cases, eg bitcoind warming up. Not sure if there is more.

src/bitcoind/interface.rs
edouardparis: To sum things up we want:
1. fail fast at startup EXCEPT when the error is the warming up error: then it must retry
2. retry only timeout, io, and too much requests errors

For the first, because at startup the retry count is set at 0, I did not handle the warming up error and I let the current behavior: https://github.com/revault/revaultd/blob/master/src/bitcoind/mod.rs#L136
but I can modify the client to loop the retry without limit for the warming up error.
What do you think ?

darosior: I think we want to:
1. At startup, never retry but when encountering a warmup error
2. During execution, retry on any non-fatal error that may occur (i think warmup is one case we need to handle here)

(1) we already do. We just need to do (2) which AFAICT is only missing the warmup special case (if i'm not missing anything.
https://github.com/revault/revaultd/blob/457ff85dc40834ccf9e78ced25547a77d7278ad5/src/bitcoind/mod.rs#L135-L142

src/bitcoind/interface.rs
```diff
@@ -89,64 +87,57 @@ impl BitcoinD {
...
+                Err(e) => match e {
+                    BitcoindError::Server(Error::Rpc(RpcError { code, .. })) => {
+                        // Retry in case of Warming up error
+                        // https://github.com/bitcoin/bitcoin/blob/dca80ffb45fcc8e6eedb6dc481d500dedab4248b/src/rpc/protocol.h#L49
+                        if code == -28 {
+                            error = Some(e);
+                        } else {
+                            return Err(e);
+                        }
+                    }
```
darosior, line 119: nit: we have a helper which does exactly this already

https://github.com/revault/revaultd/blob/457ff85dc40834ccf9e78ced25547a77d7278ad5/src/bitcoind/mod.rs#L38-L47

darosior (approved):
utACK d5bc4a0d75a392d41b7afb5302bb7e77760b988e

Thanks, it's much cleaner now. You have tested it, right?

src/bitcoind/interface.rs
darosior: It's fine as is. But i was thinking (untested):
```rust
BitcoindError::Server(_) => {
    if e.is_warming_up() {
        ...
    } else {
        ...
    }
}
```

darosior:
ACK dad2c526352134b368045e993e620c47225dcc8b

PR #397, opened 2022-05-04
Fix test_gethistory race errors

### LINKED ISSUE #396: `test_gethistory` is racy
On master you can reproduce by first removing the delays between polling `bitcoind`:
```diff
diff --git a/tests/test_framework/revaultd.py b/tests/test_framework/revaultd.py
index 1fe1ec7..008ec2f 100644
--- a/tests/test_framework/revaultd.py
+++ b/tests/test_framework/revaultd.py
@@ -76,7 +76,7 @@ class Revaultd(TailableProc):
             f.write('network = "regtest"\n')
             f.write(f"cookie_path = '{bitcoind_cookie_path}'\n")
             f.write(f"addr = '127.0.0.1:{bitcoind_rpc_port}'\n")
-            f.write("poll_interval_secs = 10\n")
+            f.write("poll_interval_secs = 0\n")
 
             if stk_config is not None:
                 f.write("[stakeholder_config]\n")
```

And running it enough times to trigger the race (shouldn't be hard to trigger with the above patch which exacerbates the race):
```
cargo build --release && REVAULTD_PATH=$PWD/target/release/revaultd POSTGRES_PASS=revault POSTGRES_USER=revault pytest -vvv --ignore tests/servers/ -k gethistory --dist each -n 10
```

edouardparis:
Rely on vaults moved_at timestamps instead of current timestamp
Generate more block to insure that the blocktime is different
for the transactions we want to test.

close #396


tests/test_rpc.py
```diff
@@ -1220,6 +1222,9 @@ def test_gethistory(revault_network, bitcoind, executor):
     events = mans[0].rpc.gethistory(["spend"], t1, t2, 20)["events"]
     assert len(events) == 0
 
+    # More block to have a clean separation in terms of blocktime
+    bitcoind.generate_block(12)
```
darosior, line 1226: Is it 12 because of the MTP rule?

darosior (approved):
ACK 42504be28bf0bdf54a13e4508d464ae2d9e9fb47

It makes sense to me that we use block times as parameter since that's what the timestamp get compared to. Mining >11 blocks to make sure the block time moved forward makes sense to me as well. I ran 30 instances of this test in parallel, with a poll_interval_secs of 1 for revaultd and got no failure.

PR #405, opened 2022-05-12
qa: split the activation of vaults for large tests

darosior:
This is causing timeouts on CI, and i'm out of optimizations: it's just
a *lot* of traffic. So, solve the issue at the root and make it wait for
smaller batches of vaults at a time.


darosior:
ACK 80701bf2214a25bef806fbc6d8016f1de0e8be20

PR #407, opened 2022-05-13
Addition & modification of instructions to run tests.

Zshan0:
The PR updates the current instructions to run revaultd tests in the following way:
- `bitcoind` requirement is implicit and is not mentioned in the README, the requirement is made explicit
- The source code already contains a script for compiling the test servers, mentioning it would make it easier for contributors & users.
- The docker requirement is mentioned in [coordinatord](https://github.com/revault/coordinatord/blob/master/CONTRIBUTING.md#running-the-tests) which also can be added here to avoid confusion.
- There is a chance that older or less powerful systems will encounter timeout errors, stating a simple way to avoid them will make debugging easier

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -19,6 +19,8 @@ python3 -m venv venv
 # Get the deps
 pip install -r tests/requirements.txt
 ```
+Additionaly you need to have `bitcoind` installed on your computer, please
+refer to [bitcoincore](https://bitcoincore.org/en/download/) for installation.
 
 ### Test modes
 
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -19,6 +19,8 @@ python3 -m venv venv
 # Get the deps
 pip install -r tests/requirements.txt
 ```
+Additionaly you need to have `bitcoind` installed on your computer, please
+refer [bitcoincore](https://bitcoincore.org/en/download/) for installation.
```
darosior, line 23: nit 
```suggestion
refer to [bitcoincore](https://bitcoincore.org/en/download/) for installation.
```

tests/README.md
```diff
@@ -57,11 +59,7 @@ must `init` the submodules (if you did not `clone` with `--recursive`) and compi
 servers code:
 ```
 # From the root of the repository
-git submodule update --init --recursive
-cd tests/servers
-cd coordinatord && cargo build && cd ..
-cd cosignerd && cargo build && cd ..
-cd miradord && cargo build && cd ..
+./contrib/recompile_tests_servers.sh
```
darosior, line 62: Just below this line it's stated:
> When you need a new version of the servers, you can update the submodules

Can you update this as well saying that they can just reuse the same script to update and recompile the tests servers?

tests/README.md
```diff
@@ -71,11 +69,19 @@ cd tests/servers
...
+```
+
 To run the server-requiring tests, pass the postgres credentials to the framework:
 ```
 POSTGRES_USER="test" POSTGRES_PASS="test" TEST_DEBUG=1 pytest -vvv -n8 --timeout=1800
 ```
 
+It is plausible that your functional tests encounter timeout error, you can 
+increase it by setting the environment variable of your shell by 
+`export TIMEOUT=120` command and run on single thread.
```
darosior, line 84: I don't think we should recommend `export`ing variables
```suggestion
It is plausible that your functional tests encounter timeout error, you can 
increase it by setting the `TIMEOUT` environment variable to a higher value
(for instance `TIMEOUT=120`). Running in single-threaded mode (by not
specifying the `-n` parameter) might also help.
```

darosior (commented):
Thanks, that's helpful!

tests/README.md
darosior: Also, compiling in release mode and setting `REVAULTD_PATH=$PWD/target/release/revaultd` is a pretty good way to avoid timeouts.

darosior (approved):
ACK 2414df911885442d047d367baa738ed5e4d9ec3e

Issue #411, opened 2022-05-17
[Milestone Tracker] Implementation of manual CPFP for Revault Managers  


Zshan0:
This issue is to update the community on the progress that will be made on the implementation of manual CPFP command for managers. The addition of a JSONRPC Command on `revaultd`’s side will be followed by the implementation of a button to do the same on `revault-gui`. The following list is for the tasks that are relevant to `revaultd` only. Please refer to my documentation [here](https://docs.google.com/document/d/1cYATCmwks22gp7fh3IEadhsxpz6sI5kwnSAg4dBnaMM/edit?usp=sharing) and feel free to drop feedbacks if necessary.
- [ ] Documenting a JSONRPC Command and its interaction with `revault-gui`
- [ ] Finalizing a design for the API
- [ ] Implementation of the API in `revaultd`
- [ ] Completion of tests for the API
- [ ] Finish the tasks related on `revault-gui`
- [ ] Final merge testing

PR #413, opened 2022-06-04
Manual CPFP

Zshan0:
This PR is to maintain code corresponding to the RFC that is being discussed for the addition of manual CPFP command as a JSONRPC API in revaultd. RFC is present [here](https://docs.google.com/document/d/1wSYSbDiSp-lk0ztzEHg4g4Ihy7BRPCG9JZE4MR6cNK0/edit?usp=sharing). #411 is the corresponding milestone tracker

### DIFF
--- a/src/bitcoind/mod.rs
+++ b/src/bitcoind/mod.rs
@@ -3,10 +3,19 @@ pub mod poller;
 pub mod utils;
 
 use crate::config::BitcoindConfig;
-use crate::{database::DatabaseError, revaultd::RevaultD, threadmessages::BitcoindMessageOut};
+use crate::{
+    commands::CommandError,
+    database::{
+        interface::{db_spend_transaction, db_unvault_transaction_by_txid},
+        DatabaseError,
+    },
+    revaultd::RevaultD,
+    threadmessages::BitcoindMessageOut,
+};
 use interface::{BitcoinD, WalletTransaction};
-use poller::poller_main;
+use poller::{poller_main, should_cpfp, ToBeCpfped};
 use revault_tx::bitcoin::{Network, Txid};
+use utils::cpfp_package;
 
 use std::{
     sync::{
@@ -187,6 +196,61 @@ fn wallet_transaction(bitcoind: &BitcoinD, txid: Txid) -> Option<WalletTransacti
         .ok()
 }
 
+fn cpfp(
+    revaultd: Arc<RwLock<RevaultD>>,
+    bitcoind: Arc<RwLock<BitcoinD>>,
+    txids: Vec<Txid>,
+    feerate: f64,
+) -> Result<Vec<Txid>, CommandError> {
+    let db_path = revaultd.read().unwrap().db_file();
+    assert!(revaultd.read().unwrap().is_manager());
+
+    let mut cpfp_txs = Vec::with_capacity(txids.len());
+    let mut cpfp_txids = Vec::with_capacity(txids.len());
+
+    // sats/vbyte -> sats/WU
+    let sats_wu = feerate / 4.0;
+
+    // sats/WU -> sats/kWU
+    let sats_kwu = (sats_wu * 1000.0) as u64;
+
+    for txid in txids.iter() {
+        let spend_tx = db_spend_transaction(&db_path, &txid).expect("Database must be available");
+
+        if let Some(unwrap_spend_tx) = spend_tx {
+            // If the transaction is of type SpendTransaction
+            let psbt = unwrap_spend_tx.psbt;
+            if should_cpfp(&bitcoind.read().unwrap(), &psbt, sats_kwu) {
+                cpfp_txs.push(ToBeCpfped::Spend(psbt));
+                cpfp_txids.push(txid.clone());
+            }
+        } else {
+            let unvault_tx = match db_unvault_transaction_by_txid(&db_path, &txid)
+                .expect("Database must be available")
+            {
+                Some(tx) => tx,
+                None => return Err(CommandError::InvalidParams("Unknown Txid.".to_string())),
+            };
+            // The transaction type is asserted to be UnvaultTransaction
+            let psbt = unvault_tx.psbt.assert_unvault();
+            if should_cpfp(&bitcoind.read().unwrap(), &psbt, sats_kwu) {
+                cpfp_txs.push(ToBeCpfped::Unvault(psbt));
+                cpfp_txids.push(txid.clone());
+            }
+        }
+    }
+
+    if cpfp_txids.len() > 0 {
+        match cpfp_package(&revaultd, &bitcoind.read().unwrap(), cpfp_txs, sats_kwu) {
+            Err(err) => return Err(CommandError::Bitcoind(err)),
+            Ok(txids) => return Ok(txids),
+        }
+    } else {
+        log::info!("Nothing to CPFP in the given list.");
+    }
+    Ok(cpfp_txids)
+}
+
 /// The bitcoind event loop.
 /// Listens for bitcoind requests (wallet / chain) and poll bitcoind every 30 seconds,
 /// updating our state accordingly.
@@ -208,7 +272,8 @@ pub fn bitcoind_main_loop(
         let _bitcoind = bitcoind.clone();
         let _sync_progress = sync_progress.clone();
         let _shutdown = shutdown.clone();
-        move || poller_main(revaultd, _bitcoind, _sync_progress, _shutdown)
+        let _revaultd = revaultd.clone();
+        move || poller_main(_revaultd, _bitcoind, _sync_progress, _shutdown)
     });
 
     for msg in rx {
@@ -250,6 +315,23 @@ pub fn bitcoind_main_loop(
                             "Sending transactions broadcast result to main thread: {}",
                             e
                         ))
+                    })?;
+            }
+            BitcoindMessageOut::CPFPTransaction(txids, feerate, resp_tx) => {
+                log::trace!("Received 'cpfptransaction' from main thread");
+
+                let _bitcoind = bitcoind.clone();
+                let _revaultd = revaultd.clone();
+                resp_tx
+                    .send(
+                        cpfp(_revaultd, _bitcoind, txids, feerate)
+                            .map(|_v| {})
+                            .map_err(|e| {
+                                BitcoindError::Custom(format!("Error CPFPing transactions: {}", e))
+                            }),
+                    )
+                    .map_err(|e| {
+                        BitcoindError::Custom(format!("Sending transaction for CPFP: {}", e))
                     })?;
             }
         }
--- a/src/bitcoind/poller.rs
+++ b/src/bitcoind/poller.rs
@@ -3,7 +3,7 @@ use crate::{
     bitcoind::{
         interface::{BitcoinD, DepositsState, SyncInfo, UnvaultsState, UtxoInfo},
         utils::{
-            cancel_txids, emer_txid, populate_deposit_cache, populate_unvaults_cache,
+            cancel_txids, cpfp_package, emer_txid, populate_deposit_cache, populate_unvaults_cache,
             presigned_transactions, unemer_txid, unvault_txid, unvault_txin_from_deposit,
             vault_deposit_utxo,
         },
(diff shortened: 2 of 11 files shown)


Zshan0:
A simple test for `cpfp` has been written, I'm not sure what other kinds of tests are required. Please let me know

src/bitcoind/mod.rs
```diff
@@ -187,6 +194,43 @@ fn wallet_transaction(bitcoind: &BitcoinD, txid: Txid) -> Option<WalletTransacti
...
+            // If the transaction is of type SpendTransaction
+            cpfp_txs.push(ToBeCpfped::Spend(unwrap_spend_tx.psbt));
+        } else {
+            // The transaction type is asserted to be UnvaultTransaction
+            let unvault_tx = db_vault_by_unvault_txid(&db_path, &txid)
+                .expect("Database must be available")
+                .unwrap()
+                .1
+                .psbt
+                .assert_unvault();
```
darosior, line 221: Hmmmm that sounds like this line may panic if you are passed a txid that does not point to anything :)

darosior (commented):
I had a look at the test for manual CPFP of multiple Spend transactions, which looks valid but cluttered.

tests/test_rpc.py
```diff
@@ -1348,3 +1348,274 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+    man.rpc.cpfp(spend_txids, 50)
+    man.wait_for_log(
+        f"CPFPed transactions with ids",
+    )
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) >= 2)
+    for spend_txid in spend_txids:
+        entry = bitcoind.rpc.getmempoolentry(spend_txid)
+        assert entry["descendantcount"] == 2
+        package_feerate = entry["fees"]["descendant"] * COIN / entry["descendantsize"]
+        assert package_feerate >= 50
```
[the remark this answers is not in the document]
Zshan0: The value is coming to be around `2.93` or so, before applying manual CPFP I'm asserting if the value is less than `50` for now

darosior:
Is this ready for review?

darosior (commented):
Just some comments on the implementation for now. I'll review the tests tomorrow.

Why do you update the `tests/servers/` submodules?

src/bitcoind/mod.rs
```diff
@@ -187,6 +194,60 @@ fn wallet_transaction(bitcoind: &BitcoinD, txid: Txid) -> Option<WalletTransacti
...
+            let psbt = unvault_tx.psbt.assert_unvault();
+            if should_cpfp(&bitcoind.read().unwrap(), &psbt, sats_kwu) {
+                cpfp_txs.push(ToBeCpfped::Unvault(psbt));
+                counter += 1;
+            }
+        }
+    }
+
+    if counter != 0 {
+        cpfp_package(&revaultd, &bitcoind.read().unwrap(), cpfp_txs, msats_wu)
```
darosior, line 244: This function expects sats/kWU. The value is actually correct but the variable name is confusingly wrong

src/bitcoind/mod.rs
```diff
@@ -187,6 +194,60 @@ fn wallet_transaction(bitcoind: &BitcoinD, txid: Txid) -> Option<WalletTransacti
...
+    //
+    // sats/vbyte -> sats/WU
+    let sats_wu = feerate / 4.0;
+    // sats/WU -> msats/WU
+    let msats_wu = (sats_wu * 1000.0) as u64;
+
+    // sats/WU -> sats/kWU
+    let sats_kwu = (sats_wu * 1000.0) as u64;
+
+    for txid in txids.iter() {
```
darosior, line 217: nit: no need for `.iter()` here, you can iter by value.

src/bitcoind/poller.rs
```diff
@@ -450,7 +450,7 @@ impl ToBeCpfped {
 // CPFP a bunch of transactions, bumping their feerate by at least `target_feerate`.
 // `target_feerate` is expressed in sat/kWU.
 // All the transactions' feerate MUST be below `target_feerate`.
-fn cpfp_package(
+pub fn cpfp_package(
```
darosior, line 453: nit: please make the docstring a doc comment (`///`) if you make this function public

src/bitcoind/poller.rs
```diff
@@ -541,7 +541,11 @@ fn cpfp_package(
 }
 
 // `target_feerate` is in sats/kWU
-fn should_cpfp(bitcoind: &BitcoinD, tx: &impl CpfpableTransaction, target_feerate: u64) -> bool {
+pub fn should_cpfp(
```
darosior, line 544: nit: please make the docstring a doc comment (`///`) if you make this function public

tests/test_rpc.py
```diff
@@ -1348,3 +1348,317 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+    man.wait_for_log("CPFPed transactions")
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == len(unvault_txids) + 1)
+    for unvault_txid in unvault_txids:
+        entry = bitcoind.rpc.getmempoolentry(unvault_txid)
+        assert entry["descendantcount"] == 2
+        package_feerate = entry["fees"]["descendant"] * COIN / entry["descendantsize"]
+        assert package_feerate >= 50
+
+
+def test_manual_cpfp_single_unvault(revault_network, bitcoind):
```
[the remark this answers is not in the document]
darosior: Also, note that the setup of each test takes some time. You should batch some related tests together to reduce runtime. For instance this test should definitely be merged with the above one.

src/bitcoind/mod.rs
```diff
@@ -252,6 +314,16 @@ pub fn bitcoind_main_loop(
...
             }
+            BitcoindMessageOut::CPFPTransaction(txids, feerate, resp_tx) => {
+                log::trace!("Received 'cpfptransaction' from main thread");
+
+                resp_tx
+                    .send(cpfp(revaultd, bitcoind, txids, feerate))
+                    .map_err(|e| {
+                        BitcoindError::Custom(format!("Sending transaction for CPFP: {}", e))
+                    })?;
+                return Ok(());
```
darosior, line 325: Heh, you copy pasted this from the ShutDown handler i guess.. But that's going to stop the bitcoind thread once you have CPFPed once. All other calls will make the daemon crash, and actually this will make it crash when stopping it as well...

darosior (commented):
The message handler in the bitcoind thread is wrong. It would return which would stop the loop, and obviously, make the program crash on shutdown or any further tentative to use the bitcoind thread (such as for another manual CPFP). I guess the reason you have so many small tests is a workaround for this bug...

tests/test_rpc.py
```diff
@@ -1348,3 +1348,317 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+            lambda: len(w.rpc.listvaults(["unvaulted"])["vaults"]) == len(vaults),
+        )
+
+    bitcoind.generate_block(CSV - 1)
+    man.wait_for_log(f"Succesfully broadcasted Spend tx '{spend_txid}'")
+
+    for w in revault_network.participants():
+        wait_for(
+            lambda: len(w.rpc.listvaults(["spending"])["vaults"]) == len(vaults),
+        )
```
darosior, line 1490: Instead of doing all this manually you should just use the corresponding "spend vaults anyhow" helper.

tests/test_rpc.py
```diff
@@ -1348,3 +1348,317 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+    # Manual CPFP trigger for spend_txid.
+    man.rpc.cpfp([spend_txid], 50)
+    man.wait_for_log(
+        f"CPFPed transactions with ids '{{{spend_txid}}}'",
+    )
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 2)
+    entry = bitcoind.rpc.getmempoolentry(spend_txid)
+    assert entry["descendantcount"] == 2
+    package_feerate = entry["fees"]["descendant"] * COIN / entry["descendantsize"]
+    assert package_feerate >= 50
```
darosior, line 1501: Maybe do the same before calling `cpfp`, to assert the feerate was lower beforehand. (Like in https://github.com/revault/revaultd/pull/413/files#r911853471)

tests/test_rpc.py
```diff
@@ -1348,3 +1348,317 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+            lambda: len(w.rpc.listvaults(["unvaulted"])["vaults"]) == len(vaults),
+        )
+
+    bitcoind.generate_block(CSV - 1)
+    man.wait_for_log(f"Succesfully broadcasted Spend tx")
+
+    for w in revault_network.participants():
+        wait_for(
+            lambda: len(w.rpc.listvaults(["spending"])["vaults"]) == len(vaults),
+        )
```
darosior, line 1539: Same here, just use the helper :)

tests/test_rpc.py
```diff
@@ -1348,3 +1348,317 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+
+    # Mining the spend and unvault transactions
+    bitcoind.generate_block(6)
+
+    # Manual CPFP trigger for already mined spend_txid.
+    man.rpc.cpfp([spend_txid], 50)
+    man.wait_for_log(
+        f"Nothing to CPFP in the given list.",
+    )
```
darosior, line 1637: This shouldn't be a log. Maybe the RPC command could return the list of the txids that were actually CPFPed? Or the txid of the CPFP transaction, so the caller could retrieve it, and in this case it would be `null`?

src/bitcoind/mod.rs
```diff
@@ -187,6 +194,60 @@ fn wallet_transaction(bitcoind: &BitcoinD, txid: Txid) -> Option<WalletTransacti
...
+            if should_cpfp(&bitcoind.read().unwrap(), &psbt, sats_kwu) {
+                cpfp_txs.push(ToBeCpfped::Spend(psbt));
+                counter += 1;
+            }
+        } else {
+            let unvault_pair =
+                db_vault_by_unvault_txid(&db_path, &txid).expect("Database must be available");
+            let unvault_tx = match unvault_pair {
+                Some((_vault, tx)) => tx,
+                None => return Err(BitcoindError::Custom("Unknown Txid.".to_string())),
```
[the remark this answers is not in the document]
Zshan0: If there is a need for `Unknown Txid` as a separate error type? we can make one, but I'm not sure if there are other scenarios where the user is allowed to query an unknown `txid`

darosior:
Untested, but i think you can use this to externally fund the CPFP wallet:
```diff
diff --git a/tests/test_framework/revault_network.py b/tests/test_framework/revault_network.py
index 0a840f8..8dfa68e 100644
--- a/tests/test_framework/revault_network.py
+++ b/tests/test_framework/revault_network.py
@@ -3,6 +3,7 @@ import logging
 import os
 import random
 
+from bip380.descriptors import Descriptor
 from ephemeral_port_reserve import reserve
 from nacl.public import PrivateKey as Curve25519Private
 from test_framework import serializations
@@ -541,6 +542,13 @@ class RevaultNetwork:
 
         return created_vaults
 
+    def fund_cpfp(self, amount):
+        """Send a coin worth this value to the CPFP descriptor"""
+        der_desc = Descriptor.from_str(str(self.cpfp_desc))
+        der_desc.derive(0)
+        decoded = self.bitcoind.rpc.decodescript(der_desc.script_pubkey.hex())
+        self.bitcoind.rpc.sendtoaddress(decoded["address"], amount)
+
     def secure_vault(self, vault):
         """Make all stakeholders share signatures for all revocation txs"""
         deposit = f"{vault['txid']}:{vault['vout']}"
```

darosior:
This should fix the unit tests CI, and also fixes the configuration of the functional tests CI (thanks for reporting it!): https://github.com/revault/revaultd/pull/423.

src/bitcoind/utils.rs
```diff
@@ -288,3 +294,93 @@ pub fn emer_txid(
...
+        package_fees += tx.fees();
+        assert!(((package_fees.as_sat() * 1000 / package_weight) as u64) < target_feerate);
+        match tx.cpfp_txin(cpfp_descriptor, &revaultd.secp_ctx) {
+            Some(txin) => txins.push(txin),
+            None => {
+                log::error!("No CPFP txin for tx '{}'", tx.txid());
+                return Ok(txids.into_iter().collect());
+            }
+        }
+    }
```
Zshan0, line 326: Rather than returning all the transactions, I need to throw an error here. I'm not sure which one

tests/test_rpc.py
```diff
@@ -1348,3 +1345,249 @@ def test_gethistory(revault_network, bitcoind, executor):
...
+    for vault in vaults:
+        deposit = f"{vault['txid']}:{vault['vout']}"
+        unvault_psbt = serializations.PSBT()
+        unvault_b64 = wallet.rpc.listpresignedtransactions([deposit])[
+            "presigned_transactions"
+        ][0]["unvault"]
+        unvault_psbt.deserialize(unvault_b64)
+        unvault_psbt.tx.calc_sha256()
+        unvault_txids.append(unvault_psbt.tx.hash)
+    return unvault_txids
```
Zshan0, line 1359: Is it a good practice to import functions from other test files? This function is present in `test_misc.py`. I can also move it into a different file. Let me know what would be the best approach

src/bitcoind/utils.rs
darosior: You can create a new enum variant for it if none other fit.

tests/test_rpc.py
darosior: I would make it a method of the `RevaultNetwork` class.

PR #415, opened 2022-06-13
`sighash_type`  made explicit along with a bump in `bitcoind` version for CI testing

### LINKED ISSUE #414: Minor API break by Bitcoin Core wallet
For the managers' CPFP we let the Bitcoin Core wallet do the work. For this we use `walletprocesspsbt` and specify the information within the PSBT. But Bitcoin Core will ignore the sighash type specified inside the PSBT inputs. So we were implicitly relying on the default value of the `sighash_type` parameter of the `walletprocesspsbt` which happened to be what we wanted (`SIGHASH_ALL`). https://github.com/bitcoin/bitcoin/pull/22514 changed the default to `SIGHASH_DEFAULT`, which breaks our usage.

The fix is trivial, and backward compatible with previous versions of bitcoind. We need to explicitly specify the `sighash_type` parameter when calling `walletprocesspsbt`:
https://github.com/revault/revaultd/blob/076df0d79e61df92e9f2f0b55ae7a12a1e46a58b/src/bitcoind/interface.rs#L722-L746

Zshan0:
Since the Bitcoin Core ignores the sighash type specified inside the PSBT inputs, `sign_psbt` relied on the default value of `sighash_type` which aligned with our requirements. https://github.com/bitcoin/bitcoin/pull/22514 changed the default value from `SIGHASH_ALL` to `SIGHASH_DEFAULT`. Making `sighash_type` explicit fixes the minor breaking change caused by the update in the bitcoin wallet. Note that since `sighashtype` parameter comes after the `sign` parameter, the value of `sign` has been set to the default value, `true`. 

Along with the fix, `bitcoind` version used in CI has been bumped from `22.0` to `23.0`. 

This issue closes #414.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -9,7 +9,7 @@ task:
     VERBOSE: 0
     LOG_LEVEL: debug
     TIMEOUT: 300
-    BITCOIND_VERSION: 22.0
+    BITCOIND_VERSION: 23.0
     BITCOIND_DIR_NAME: "bitcoin-$BITCOIND_VERSION"
   matrix:
     - name: 'Misc functional tests'
--- a/src/bitcoind/interface.rs
+++ b/src/bitcoind/interface.rs
@@ -726,7 +726,11 @@ impl BitcoinD {
     pub fn sign_psbt(&self, psbt: &Psbt) -> Result<(bool, Psbt), BitcoindError> {
         let res = self.make_cpfp_request(
             "walletprocesspsbt",
-            &params!(Json::String(base64::encode(&encode::serialize(psbt)))),
+            &params!(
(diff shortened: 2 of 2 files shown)


Zshan0:
Keeping the `sighash_type` as `SigHashType::All` was throwing an error:
```
Error while updating tip: \'Bitcoind server error: RPC error response: RpcError { code: -1, message: "SIGHASH_ALL is not a valid sighash parameter.", data: None }\''
```
After looking at [cljdocs](https://cljdoc.org/d/bitcoin-rpc/bitcoin-rpc/0.2.3/api/bitcoin-rpc.core) I changed the value to `"ALL"` and the tests passed locally. I was not able to find the correct Macro for it, please let me know if there are any more changes required.

PR #417, opened 2022-06-30
bitcoind: fix misuse of listsinceblock filtering by block hash

darosior:
I had encountered this too when switching the poller to use `listsinceblock`.

But this time i've upstreamed a modification of the doc so this doesn't happen a third time. :) https://github.com/bitcoin/bitcoin/pull/25510


darosior:
Note that the code was technically correct before this patch since it's used to detect the Unvault spender, and the Unvault Script always has a CSV of at least 1. Still, better for the helper to be correct.

edouardparis (approved):
ACK b216e7db829f2b5ff566c68316c1aae7dc3db9af

